PR #2, opened 2020-06-19
[CA-869] Basic Janitor service setup

yonghaoy:
This includes:

1. Basic build.gradle
2: openapi
3: readme
4: Basic spring-boot
5: two endpoints - status and getResourceId. 
I am having two endpoints because the status is the real one and will be used. getResourceId is a fake one as an example of some openapi use cases


README.md
```diff
@@ -1,2 +1,15 @@
...
+## Using the Gradle wrapper
+Set executable permissions:
+```
+chmod +x gradlew
+```
+Start local server
+```
+./gradlew bootRun
+```
+And then check http://127.0.0.1:8080/status for service status
```
wchamber, line 15: maybe add the swagger url as well if it's working?

README.md
```diff
@@ -1,2 +1,15 @@
...
+# Development
+
+## Using the Gradle wrapper
+Set executable permissions:
+```
+chmod +x gradlew
+```
+Start local server
+```
+./gradlew bootRun
```
wchamber, line 13: I think we're missing the `gradle/wrapper/gradle-wrapper.jar`. Do you have it locally only?

yonghaoy: oh right, I ignore .jar files because I saw that is what trd did:
https://github.com/DataBiosphere/jade-data-repo/blob/develop/.gitignore#L17.
Added it back to something like https://github.com/DataBiosphere/terra-workspace-manager/blob/master/.gitignore#L5

yonghaoy: created https://broadworkbench.atlassian.net/browse/CA-909

PR #3, opened 2020-06-23
[CA-870] add github action to deploy crl-janitor

yonghaoy:
The push file is copied from kernel-service-poc:
https://github.com/DataBiosphere/kernel-service-poc/blob/master/.github/workflows/master_push.yml


.github/workflows/master_push.yml
```diff
@@ -0,0 +1,81 @@
...
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - 'README.md'
+      - '.github/**'
+env:
+  SERVICE_NAME: ${{ github.event.repository.name }}
+  GOOGLE_PROJECT: terra-kernel-k8s
+  GKE_CLUSTER: terra-kernel-k8s
```
andy7i, line 13: these look like they're the same as the ones in kernel-service-poc. Is it ok for them to be the same or did we want them to be crl-janitor specific?

yonghaoy: terra-kernel-k8s is the GKE cluster to host all MC Terra components. Now it has WM, SAM, IC, POC. I think for crl-janitor, it can also be in this cluster along with other services.

PR #6, opened 2020-06-25
[CA-871] Initial DB setup using liquibase

yonghaoy:
Ticket: https://broadworkbench.atlassian.net/browse/CA-871

The initial schema comes from this doc(might change in the future): 
https://docs.google.com/document/d/13mYVJML_fOLsX1gxQxRJgECUNT27dAMKzEJxUEvtQqM/edit#heading=h.8pzrjwy9ihf9

I learnt the code structure from WM's code: 
https://github.com/DataBiosphere/terra-workspace-manager/tree/master/src/main/java/bio/terra/workspace/app

crljanitor=> \dt
                  List of relations
 Schema |         Name          | Type  |   Owner    
--------+-----------------------+-------+------------
 public | cleanup_log           | table | crljanitor
 public | databasechangelog     | table | crljanitor
 public | databasechangeloglock | table | crljanitor
 public | gcp_bucket            | table | crljanitor
 public | label                 | table | crljanitor
 public | resource_type         | table | crljanitor
 public | tracked_resource      | table | crljanitor

### DIFF
--- a/src/main/java/bio/terra/janitor/app/StartupInitializer.java
+++ b/src/main/java/bio/terra/janitor/app/StartupInitializer.java
@@ -1,0 +1,25 @@
+package bio.terra.janitor.app;
+
+import bio.terra.janitor.app.configuration.JanitorJdbcConfiguration;
+import bio.terra.janitor.service.migirate.MigrateService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationContext;
+
+public final class StartupInitializer {
+  private static final Logger logger = LoggerFactory.getLogger(StartupInitializer.class);
+  private static final String changelogPath = "db/changelog.xml";
+
+  public static void initialize(ApplicationContext applicationContext) {
+    // Initialize or upgrade the database depending on the configuration
+    MigrateService migrateService = (MigrateService) applicationContext.getBean("migrateService");
+    JanitorJdbcConfiguration janitorJdbcConfiguration =
+        (JanitorJdbcConfiguration) applicationContext.getBean("janitorJdbcConfiguration");
+
+    if (janitorJdbcConfiguration.isInitializeOnStart()) {
+      migrateService.initialize(changelogPath, janitorJdbcConfiguration.getDataSource());
+    } else if (janitorJdbcConfiguration.isUpgradeOnStart()) {
+      migrateService.upgrade(changelogPath, janitorJdbcConfiguration.getDataSource());
+    }
+  }
+}
--- a/src/main/java/bio/terra/janitor/app/StartupInitializer.java
+++ b/src/main/java/bio/terra/janitor/app/StartupInitializer.java
@@ -6,8 +6,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
 
+/**
+ * Initializes the application after the application is setup, but before the port is opened for business.
+ * The purpose for this class is for database initialization and migration.
+ */
 public final class StartupInitializer {
-  private static final Logger logger = LoggerFactory.getLogger(StartupInitializer.class);
   private static final String changelogPath = "db/changelog.xml";
 
   public static void initialize(ApplicationContext applicationContext) {
--- a/src/main/java/bio/terra/janitor/app/StartupInitializer.java
+++ b/src/main/java/bio/terra/janitor/app/StartupInitializer.java
@@ -16,9 +16,9 @@ public final class StartupInitializer {
     JanitorJdbcConfiguration janitorJdbcConfiguration =
         (JanitorJdbcConfiguration) applicationContext.getBean("janitorJdbcConfiguration");
 
-    if (janitorJdbcConfiguration.isInitializeOnStart()) {
+    if (janitorJdbcConfiguration.isRecreateDbOnStart()) {
       migrateService.initialize(changelogPath, janitorJdbcConfiguration.getDataSource());
-    } else if (janitorJdbcConfiguration.isUpgradeOnStart()) {
+    } else if (janitorJdbcConfiguration.isUpdateDbOnStart()) {
       migrateService.upgrade(changelogPath, janitorJdbcConfiguration.getDataSource());
     }
   }
--- a/src/main/java/bio/terra/janitor/app/configuration/JanitorJdbcConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/JanitorJdbcConfiguration.java
@@ -18,23 +18,23 @@ public class JanitorJdbcConfiguration extends JdbcConfiguration {
   // 1. initialize is true; upgrade is irrelevant - initialize and recreate an empty database
   // 2. initialize is false; upgrade is true - apply changesets to an existing database
   // 3. initialize is false; upgrade is false - do nothing to the database
-  private boolean initializeOnStart;
-  private boolean upgradeOnStart;
+  private boolean recreateDbOnStart;
+  private boolean updateDbOnStart;
 
-  public boolean isInitializeOnStart() {
-    return initializeOnStart;
+  public boolean isRecreateDbOnStart() {
+    return recreateDbOnStart;
   }
 
-  public void setInitializeOnStart(boolean initializeOnStart) {
-    this.initializeOnStart = initializeOnStart;
+  public void setRecreateDbOnStart(boolean recreateDbOnStart) {
+    this.recreateDbOnStart = recreateDbOnStart;
   }
 
-  public boolean isUpgradeOnStart() {
-    return upgradeOnStart;
+  public boolean isUpdateDbOnStart() {
+    return updateDbOnStart;
   }
 
-  public void setUpgradeOnStart(boolean upgradeOnStart) {
-    this.upgradeOnStart = upgradeOnStart;
+  public void setUpdateDbOnStart(boolean updateDbOnStart) {
+    this.updateDbOnStart = updateDbOnStart;
   }
 
   // This bean plus the @EnableTransactionManagement annotation above enables the use of the
--- a/src/main/java/bio/terra/janitor/app/configuration/JanitorJdbcConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/JanitorJdbcConfiguration.java
@@ -15,9 +15,9 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
 public class JanitorJdbcConfiguration extends JdbcConfiguration {
   // These properties control code in the StartupInitializer. We would not use these in production,
   // but they are handy to set for development and testing. There are only three interesting states:
-  // 1. initialize is true; upgrade is irrelevant - initialize and recreate an empty database
-  // 2. initialize is false; upgrade is true - apply changesets to an existing database
-  // 3. initialize is false; upgrade is false - do nothing to the database
+  // 1. recreateDbOnStart is true; updateDbOnStart is irrelevant - initialize and recreate an empty database
+  // 2. recreateDbOnStart is false; updateDbOnStart is true - apply changesets to an existing database
+  // 3. recreateDbOnStart is false; updateDbOnStart is false - do nothing to the database
   private boolean recreateDbOnStart;
   private boolean updateDbOnStart;
 
--- a/src/main/java/bio/terra/janitor/app/configuration/JanitorJdbcConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/JanitorJdbcConfiguration.java
@@ -1,0 +1,46 @@
+package bio.terra.janitor.app.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.stereotype.Component;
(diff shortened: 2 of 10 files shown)


src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+              constraints:
+                references: tracked_resource(tracked_id)
+                foreignKeyName: fk_tracked_id
+                primaryKey: true
+                nullable: false
+          - column:
+              name: key
+              type: text
+              constraints:
+                nullable: true
```
gpolumbo-broad, line 84:
```suggestion
                nullable: false
```

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+              constraints:
+                references: tracked_resource(tracked_id)
+                foreignKeyName: fk_tracked_id
+                primaryKey: true
+                nullable: false
+          - column:
+              name: logtime
+              type: timestamp
+              constraints:
+                nullable: true
```
gpolumbo-broad, line 107:
```suggestion
                nullable: false
```

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+                unique: false
+          - column:
+              name: status
+              type: text
+              constraints:
+                nullable: true
```
gpolumbo-broad, line 113:
```suggestion
                nullable: false
```

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+                unique: false
+          - column:
+              name: description
+              type: text
+              constraints:
+                nullable: true
```
gpolumbo-broad, line 119: Maybe?  My preference would be that all log entries require a description, but I wouldn't be surprised if we find some obscure scenarios where it is not needed?

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+                unique: false
+    - createTable:
+        tableName: gcp_bucket
```
gpolumbo-broad, line 122: Slightly concerned that this might indicate a proliferation of tables for different resource types across different providers, which feels like we might be doing something wrong.  I would feel a little better if it were just a `buckets` table that could store information for all cloud providers in one table, but even that I'd like to avoid if we can.

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+              constraints:
+                references: tracked_resource(tracked_id)
+                foreignKeyName: fk_tracked_id
+                primaryKey: true
+                nullable: false
+          - column:
+              name: project_id
+              type: text
+              constraints:
+                nullable: true
```
gpolumbo-broad, line 136:
```suggestion
                nullable: false
```

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+                unique: false
+          - column:
+              name: bucket_name
+              type: text
+              constraints:
+                nullable: true
```
gpolumbo-broad, line 142:
```suggestion
                nullable: false
```

src/main/java/bio/terra/janitor/service/migirate/MigrateService.java
```diff
@@ -0,0 +1,83 @@
...
+   * database schema is in a valid state for applying changesets to.
+   *
+   * @param changesetFile - relative path to the changeset file in the project
+   * @param dataSource - database to operate on
+   */
+  public void upgrade(String changesetFile, DataSource dataSource) {
+    migrateWorker(changesetFile, dataSource, false);
+  }
+
+  private void migrateWorker(String changesetFile, DataSource dataSource, boolean initialize) {
```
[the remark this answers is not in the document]
yonghaoy: ah, that make sense. changed to destroyAll

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+databaseChangeLog:
+- changeSet:
+    id: init_janitor_table
+    author: yonghao
+    changes:
+    - createTable:
+        tableName: resource_type
+        columns:
+          - column:
+              name: type_id
```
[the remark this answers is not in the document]
yonghaoy: changed for resource_type and tracked_resource table.
For gcp_bucket, cleanup_log, label table, still use tracked_id as I feel that might be better since whose are the FK from tracked_resource, and tracker_id is easier to understand?

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+              type: uuid
+              constraints:
+                primaryKey: true
+                nullable: false
+          - column:
+              name: name
+              type: text
+              constraints:
+                nullable: true
```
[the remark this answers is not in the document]
yonghaoy: changed all to true

src/main/resources/db/changesets/20200625_initial_schema.yaml
yonghaoy: I think this is optional and should be nullable?

src/main/resources/db/changesets/20200625_initial_schema.yaml
yonghaoy: Right, the gcp_bucket does sound like a cloud resource type, but I feel the current schema looks like a gcp specific thing: (project_id + bucket_name)? I feel this table may need some future discussion. I will check with Will once he is back.

Is there any better table name which includes "gcp specific" + "bucket" but doesn't sound like a resource type?

wchamber: So the gcp_bucket table was meant as a dummy example of a database schema where we did do a proliferation of tables for different resource types. The other option proposed in the doc was the `resource_uid` json field. We should not do both. We did not discuss this in any review of the doc: https://docs.google.com/document/d/13mYVJML_fOLsX1gxQxRJgECUNT27dAMKzEJxUEvtQqM/edit#heading=h.vhh24t1ac0cs
So options:
1. json field to store arbitrary cloud resource uid on `tracked_resource` table. 
2. proliferation of tables, one per resource type with nasty joins across many tables.
Third option wasn't in the doc, but is being done here:
3. table with union of all fields from all resource types, as is being started here. Every field can be nullable, we keep extending this table whenever we need to add a resource type with a new column name.
I think we're better off discussing this offline then in the comments here.

src/main/java/bio/terra/janitor/app/configuration/JanitorJdbcConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+@Component
+@EnableConfigurationProperties
+@EnableTransactionManagement
+@ConfigurationProperties(prefix = "db.janitor")
+public class JanitorJdbcConfiguration extends JdbcConfiguration {
+  // These properties control code in the StartupInitializer. We would not use these in production,
+  // but they are handy to set for development and testing. There are only three interesting states:
+  // 1. initialize is true; upgrade is irrelevant - initialize and recreate an empty database
+  // 2. initialize is false; upgrade is true - apply changesets to an existing database
+  // 3. initialize is false; upgrade is false - do nothing to the database
```
wchamber, line 20: comments need to be updated to match variable names.

src/main/java/bio/terra/janitor/service/migirate/MigrateService.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * The MigrateService invokes Liquibase to configure the service database. It provides the methods
+ * to clear stale locks, drop all data, and update the service database. We implement this in code,
+ * as opposed to using hooks in the Spring framework. We expect MC Terra services to scale
+ * horizontally and to perform rolling upgrades of the running instances. Therefore, we do not want
+ * the act of running an instance to suddenly upgrade the database schema.
+ */
+@Component
+public class MigrateService {
```
wchamber, line 24: This feels like it should go in some common terra library repo. but I guess let's do that later.

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+          - column:
+              name: resource_type_id
+              type: uuid
+              constraints:
+                references: resource_type(id)
+                foreignKeyName: fk_type_id
+                nullable: false
+          - column:
+              name: resource_uid
+              type: json
```
wchamber, line 41: I think we want `jsonb`. https://www.postgresql.org/docs/9.4/datatype-json.html

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+              constraints:
+                references: resource_type(type_id)
+                foreignKeyName: fk_type_id
+                nullable: true
+          - column:
+              name: resource_uid
+              type: json
+              constraints:
+                nullable: true
+                unique: false
```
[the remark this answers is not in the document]
wchamber: I don't think we want uniqueness here. It's valid that we track the deletion of multiple resources that have the same cloud resource uid across time. e.g. I create project `gcp_projcet: "my-project"`, janitor deletes it. a week later, I make the same project for the janitor to delete. The janitor will still have the record of the first resource with it makred as cleaned up.

There's a good logic question we discussed of how we want to "de-dupe" multiple tracked resources with the same cloud resource uid - e.g. they refer to the same cloud resource. I don't think we need to answer that for the database schema.

src/main/resources/db/changesets/20200625_initial_schema.yaml
wchamber: I think we'd end up writing dummy descriptions that might as well be null. So, I agree that nullable is fine.
Like, what's the description for when the resource is already cleaned up? That's all captured in the status.

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -0,0 +1,143 @@
...
+                nullable: false
+          - column:
+              name: logtime
+              type: timestamp
+              constraints:
+                nullable: false
+                unique: false
+          - column:
+              name: status
+              type: text
```
wchamber, line 111: Do we want to use an enum? Ditto for `tracked_resource.state`. I defer to others with practical db experience.
https://www.postgresql.org/docs/9.1/datatype-enum.html

src/main/java/bio/terra/janitor/app/StartupInitializer.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.janitor.app;
+
+import bio.terra.janitor.app.configuration.JanitorJdbcConfiguration;
+import bio.terra.janitor.service.migirate.MigrateService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationContext;
+
+public final class StartupInitializer {
```
wchamber, line 9: I'd appreciate a comment on what this class is for and how/when it should be used in spring configuration.

src/main/resources/db/changesets/20200625_initial_schema.yaml
yonghaoy: Option 2 sounds too many tables will be needed and I'd prefer option 1 looks good to me.  
Reason: seems json query vs creates/maintains 10+tables json query is simpler and easier. A nice json query template is easier to implement (https://www.postgresqltutorial.com/postgresql-json/) compare with a template for updating/inserting/query across tables based on resource type? 

this pr has 1 and 2 which is not correct. Do we want this initial schema just have the 1st option(or 2 or none) for now and change later once we decide. Or do we want to hold on this pr and wait until decision being made?

wchamber: I vote for option 1 and let this PR move ahead. I prefer 1 to 2 or 3. If we change our minds, at some point in development, we can change our database.

src/main/java/bio/terra/janitor/app/StartupInitializer.java
yonghaoy: added

src/main/resources/db/changesets/20200625_initial_schema.yaml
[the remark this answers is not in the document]
gpolumbo-broad: Ok, sounds good to me

src/main/resources/db/changesets/20200625_initial_schema.yaml
gpolumbo-broad: Historically, I have not used this.  We would either just use a `TEXT` field and maintain the enum in the code base, or extract the list of states into their own table and link them with a foreign key.  I think the issue ultimately boiled down to not wanting to maintain the enum in the code and the database and keep them in sync.  

That said, I don't really have a strong preference for whatever you choose to do here, I think it might be one of those situations where you can try it out and see how it works and you'll probably know pretty quickly how good or bad it looks.

gpolumbo-broad:
@wchamber - as to the `resource_type` as enum question, my gut instinct on that one is to say no, keep it a table.  That feels like a something best represented as a full model entity that can have additional attributes appended to it as need be, it's more than just a controlled list of strings.

PR #7, opened 2020-06-26
[CA 918] setup local development using skaffold

yonghaoy:
Branched from: https://github.com/DataBiosphere/terra-workspace-manager/tree/master/local-dev

The Janitor service:
https://console.cloud.google.com/kubernetes/service/us-central1-a/terra-integration/terra-yyu/terra-crl-janitor-service/overview?organizationId=548622027621&project=terra-kernel-k8s


local-dev/setup_local_env.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Clone Helm chart and helmfile repos
+rm -rf terra-helm
+rm -rf terra-helmfile
+git clone -b "$TERRA_HELM_BRANCH" --single-branch https://github.com/broadinstitute/terra-helm
+git clone -b "$TERRA_HELMFILE_BRANCH" --single-branch https://github.com/broadinstitute/terra-helmfile
+
+# Template in environment
+sed "s|ENV|${ENV}|g" skaffold.yaml.template > skaffold.yaml
+sed "s|ENV|${ENV}|g" values.yaml.template > values.yaml
```
wchamber, line 23: yuck. But I guess this is what we're doing for templating for now.
Maybe at least make the replaced string in the yaml more unique?

local-dev/skaffold.yaml.template
```diff
@@ -0,0 +1,22 @@
...
+build:
+  artifacts:
+  - image: gcr.io/terra-kernel-k8s/terra-crl-janitor
+    context: ../
+    jib: {}
+deploy:
+  helm:
+    releases:
+      - name: crljanitor-ENV
+        namespace: terra-ENV
```
wchamber, line 13: Can we use the helm values instead here and then only sed replace the values.yaml.template (and make this `skaffold.yaml` instead of a template?
`-name: crljanitor-{{ .Values.global.terraEnv }}`

local-dev/local-postgres-init.sql
```diff
@@ -0,0 +1,4 @@
+CREATE DATABASE testdb;
+CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
+CREATE DATABASE janitor;
+CREATE ROLE janitoruser WITH LOGIN ENCRYPTED PASSWORD 'janitorpwd';
```
wchamber, line 4: is this file used yet? I don't think we're running postgres locally with skaffold? Next iteration for local db instead of cloud sql?

local-dev/setup_local_env.sh
yonghaoy: sorry, not quite understand what you mean here. What we replace here is the env name, e.g., we have terra-yyu namespace in our k8s cluster. And what it does here by running `./setup_local_env.sh yyu`, it will generate the yaml which can be deployed to terra-yyu namespace. So the replaced string has to be "yyu", what you mean "make the replaced string in the yaml more unique"?

local-dev/local-postgres-init.sql
yonghaoy: right, we don't need this yet. Will make a separate PR when we need local db

local-dev/skaffold.yaml.template
gmalkov: Nope, Skaffold doesn't have that context

local-dev/setup_local_env.sh
wchamber: I mean the `ENV` it's replacing in the .template files.
So here, `sed "s|MY_SPECIAL_ENV|${ENV}|g" skaffold.yaml.template > skaffold.yaml` and in the .template files replace `ENV` with `MY_SPECIAL_ENV`
 I don't know what a good replacement name is, since 'ENV' is very generic to start. idk, `DEV_ENV`, `DEV_NAMESPACE`, open to anything.

local-dev/skaffold.yaml.template
wchamber: Kind of unfortunate. My 1 min uninformed read is there's some integration? https://skaffold.dev/docs/pipeline-stages/deployers/helm/
Hopefully it doesn't matter much for local configuration, but it's yet another configuration template variation with ad-hoc sed tooling.

local-dev/setup_local_env.sh
yonghaoy: "env" is what we used in terra-helm and helmfile:
https://github.com/broadinstitute/terra-helmfile/blob/master/helmfile.yaml#L10
Use env here because it does the corresponding terra env deployment. Use env here makes it clear what this does and how that works with helmfile?  What do you think? Do you want NAMESPACE instead?

wchamber: No,let's just go with ENV.

PR #8, opened 2020-07-01
[CA 871] set up local postgres db and rearrange some file structure

yonghaoy:


local-dev/run_local.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+psql -f local-dev/local-postgres-init.sql
```
wchamber, line 2: Isn't this just using/requiring you have psql running locally with default access? I see that's what data-repo and workspace manager are doing, but running postgres with docker like Sam seems like a better idea. 
Maybe something like https://github.com/broadinstitute/sam/blob/develop/docker/run-postgres.sh ?

yonghaoy: Checked with Zarc about that. The reason we have the local postgres running is to have the local version of WM with the data. And that is why they have two db, testdb and wmdb. 
Testdb is for unit test and they will reset the db when server starts.
 But for the wm db, that is for local running, and data will not be reset by default. So whenever you need the local WM, you starts it and connect to the db in your local, you will have all the data. 
The SAM's one is better if we don't want to keep those databases and data? Let me change to that one. Thanks

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+# The CloudSQL console simply states "PostgreSQL 9.6" so we may not match the minor version number
+POSTGRES_VERSION=12.3
```
wchamber, line 4: 9.6 and 12.3 seem like more than minor major version number.

README.md
```diff
@@ -3,15 +3,40 @@ Janitor service to cleanup resources created by [Cloud Resource Library (CRL)](h
...
 ```
 And then check http://127.0.0.1:8080/status for service status.
 
-Swagger in local server: http://127.0.0.1:8080/swagger-ui.html
\ No newline at end of file
+Swagger in local server: http://127.0.0.1:8080/swagger-ui.html
+
+You can connect to local postgres by running: 
+```
+psql postgresql://127.0.0.1:5432/testdb -U janitor-test
```
wchamber, line 26: is this still right?

local-dev/sql_validate.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+SERVICE=$1
+
+# validate mysql
+echo "sleeping for 5 seconds during postgres boot..."
+sleep 5
+export PGPASSWORD=${SERVICE}-test
+psql --username ${SERVICE}-test --no-password --host=postgres --port=5432 -d testdb -c "SELECT VERSION();SELECT NOW()"
```
wchamber, line 9: why doesn't this need a password? maybe get rid of the SERVICE=$1 since it's always the janitor?

local-dev/run_postgres.sh
yonghaoy: 9.6 is too old. The verson on GCP is 12 now. So I just choose the latest verison.

README.md
yonghaoy: yes, it still works.

local-dev/run_postgres.sh
wchamber: Maybe just remove the comment on line 3 then?

local-dev/sql_validate.sh
yonghaoy: I think this is no password promot. 
https://www.postgresql.org/docs/12/app-psql.html
The actually password is set by the PGPASSWORD env variable

local-dev/run_postgres.sh
yonghaoy: done

yonghaoy:
Thanks for the review. 
I do have a ticket and plan to adding test(probably go with the status endpoint to include the postgres healthy work) planned. But not in this PR.

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "Usage: $0 stop|start"
+    exit 1
+fi
+
+if [ $COMMAND = "start" ]; then
+    start
+elif [ $COMMAND = "stop" ]; then
+    stop
+else
+    exit 1
```
wchamber, line 46: i think you're missing a trailing `fi` to close the `else`. for  stop to work.

PR #9, opened 2020-07-01
[CA 909] Add a test to assert service and db healthy

yonghaoy:


.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          java-version: 1.8
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
```
wchamber, line 43: just curious, where'd you find this?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Run unit tests
+        run: ./gradlew unitTest --scan
+      - name: Stop Postgres DB
+        run:  ./local-dev/run_postgres.sh stop
```
wchamber, line 49: do we need to stop the database? Maybe keeping this just to test that `run_postgres.sh stop` works is a good reason to have it.
if so, we probably still want it to run even if the unit tests fail, as long as the initialize database step has succeeded? see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif

README.md
```diff
@@ -44,4 +44,22 @@ You may need to use gcloud to provide GCR
 
 ```
 local-dev/setup_local_env.sh <environment>
+```
+
+## Testing
```
wchamber, line 49: hmm, should this go in the local_dev README instead? don't have a strong opinion.

src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
```diff
@@ -2,14 +2,43 @@
...
+
+  @Autowired
+  UnauthenticatedApiController(JanitorJdbcConfiguration jdbcConfiguration) {
+    this.jdbcTemplate = new NamedParameterJdbcTemplate(jdbcConfiguration.getDataSource());
+  }
+
   @Override
   public ResponseEntity<SystemStatus> serviceStatus() {
-    return new ResponseEntity<>(new SystemStatus().ok(true), HttpStatus.OK);
+    if (jdbcTemplate.getJdbcTemplate().execute(this::isConnectionValid)) {
```
wchamber, line 26: wdyt about inlining `isConnectionValid`? don't think we're getting much readability benefit from having a named function.

`.execute(connection - > connection.isValid(0)) {...`

src/test/java/bio/terra/janitor/app/ServerStartUpTest.java
```diff
@@ -0,0 +1,3 @@
+package bio.terra.janitor.app;
+
+public class ServerStartUpTest {}
```
wchamber, line 3: what's up with this file?

.github/workflows/unit_test.yml
yonghaoy: ah, I copied this from: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/.github/workflows/unit-test.yml

.github/workflows/unit_test.yml
yonghaoy: ah, sorry, forgot to delete this line. That is my previous version: Manually start and stop ad postgres in docker. 
Now we are using postgres service container (line 13-24) so we don't need to manually start and stop it.

README.md
yonghaoy: Feels the WM way has to many readmes and hard to follow. So I am thinking about to have one readme and will have local server, deploy to gke, how to run test sections. Removed local_dev/ readme and put everythinh here.

src/test/java/bio/terra/janitor/app/ServerStartUpTest.java
yonghaoy: deleted this file

src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
yonghaoy: done

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,47 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Run Unit Tests
+
+on: [push]
```
wchamber, line 6: Should this run on pull requests as well?
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/7abed537fa4a6d02189dde8b78d62aff6ebd100d/.github/workflows/build-and-test.yaml#L6-L10

PR #10, opened 2020-07-07
[CA 873] [Schema] Endpoint for post tracked resource 

yonghaoy:
Seems there are some schema question worth to be discussed before we start the actual implementation, so sent the split the ticket to multiple PR


.gitignore
```diff
@@ -42,6 +42,7 @@ out/
 ### OpenAPI CodeGen detritus
 .openapi-generator/
 pom.xml
+src/main/resources/api/cloud_resources_uid
```
wchamber, line 45: This seems strange? I'd hope that gradle doesn't materialize our yaml into src/main. (something under `build` would be better). Or was this for dev while you were getting packaging fixed?

build.gradle
```diff
@@ -117,4 +121,4 @@ task unitTest(type: Test) {
         events = ["passed", "failed", "skipped", "started"]
     }
     outputs.upToDateWhen { false }
-}
+}
```
wchamber, line 124: nit: keep the end of file newline

src/main/java/bio/terra/janitor/app/configuration/ApiResourceConfig.java
```diff
@@ -13,5 +13,6 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
         .addResourceHandler("/swagger-webjar/**")
         .addResourceLocations("classpath:/META-INF/resources/webjars/swagger-ui/3.25.4/");
     registry.addResourceHandler("/**").addResourceLocations("classpath:/api/");
+    // GoogleBigQueryTableUid u = null;
```
wchamber, line 16: leftover?

src/main/resources/api/service_openapi.yaml
```diff
@@ -24,6 +24,37 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/CreateResourceRequestBody'
+      responses:
+        200:
+          description: Ok
+          $ref: '#/components/responses/CreateResponse'
+        400:
+          description: Bad request - invalid id, badly formed
+          $ref: '#/components/responses/ErrorResponse'
+        401:
```
wchamber, line 51: I think 403 is more appropriate. 401 is more like we weren't able to authenticated correctly, 403 for authz.

src/main/resources/api/service_openapi.yaml
```diff
@@ -24,6 +24,37 @@ paths:
...
+          description: Permission denied
+          $ref: '#/components/responses/ErrorResponse'
+        500:
+          description: Creation request error
+          $ref: '#/components/responses/ErrorResponse'
+
   '/api/janitor/v1/resource/{id}':
```
wchamber, line 58: maybe add a `// TODO remove temporrary placeholder endpoint.` to make it clear what's permanent vs temporary.

src/main/resources/api/service_openapi.yaml
```diff
@@ -50,6 +81,16 @@ components:
...
         type: string
+
+    Type:
+      name: type
+      in: path
+      description: The reousrce type
+      required: true
+      schema:
+        type: string
+        enum: [google-bigquery-dataset, google-bigquery-table, google-blob, google-bucket, google-project]
```
wchamber, line 92: I don't think we need this. As long as we use a CloudResourceUid schema object in the request, we can infer the resource type from that.

src/main/resources/api/service_openapi.yaml
```diff
@@ -79,6 +121,34 @@ components:
...
+          type: object
+        labels:
+          description: The labels for the resource
+          type: object
+          additionalProperties:
+            type: string
+        timeToLive:
+          description: How long created resources should live (at least) before being cleaned up.
+          type: integer
+          format: int64
```
wchamber, line 139: this needs a unit. Is it minutes? timeToLiveInMinutes?

src/main/resources/api/service_openapi.yaml
```diff
@@ -79,6 +121,34 @@ components:
                 type: array
                 items:
                   type: string
+
+    CreateResourceRequestBody:
+      type: object
+      properties:
+        resourceUid:
+          description: The UID of the resource in CRL format
+          type: object
```
wchamber, line 130: this should use our cloud_resource_uid.yaml CloudResourceUid object. Or are you planning on adding that later?

src/main/resources/api/service_openapi.yaml
```diff
@@ -79,6 +121,34 @@ components:
...
+          additionalProperties:
+            type: string
+        timeToLive:
+          description: How long created resources should live (at least) before being cleaned up.
+          type: integer
+          format: int64
+        creationTime:
+          description: When the resource is created.
+          type: integer
+          format: int64
```
wchamber, line 143: this needs units as well. Is it minutes since the epoch, seconds since the epoch?

.gitignore
yonghaoy: ah, that is the left over. Previously my build.gradle will unzip CloudResourceSchema package and copied it here then delete it. Now seems we don't need this, so that is the leftover.
On the other hand, if I unzip this under `build`, my service_openapi.yaml can not find that file, so I put this under src/

src/main/resources/api/service_openapi.yaml
yonghaoy: I see. I copied this from WM, they use 401, so I followed their pattern

src/main/resources/api/service_openapi.yaml
yonghaoy: So it would be CloudResourceUid isinstanceof GoogleBucketUID in the code?

src/main/resources/api/service_openapi.yaml
yonghaoy: Like what I said in previous comments, seems we don't need to depend on CloudResourceUid.yaml. We will parse and convert this in the code?

src/main/resources/api/service_openapi.yaml
yonghaoy: A timestamp in mills?

src/main/resources/api/service_openapi.yaml
wchamber: CloudResourceUid in Java is a super class of GoogleBucketUid. So when we get the CloudResourceUid from the request, we can do a series of if `instanceOf GoogleBucketUid` to cast to the right subclass.

src/main/resources/api/service_openapi.yaml
wchamber: I think we should depend on it. Then our schema says what it should be and in code we get the CloudResourceUid from the request. Maybe we should sync offline.

src/main/resources/api/service_openapi.yaml
```diff
@@ -24,6 +24,37 @@ paths:
           description: Service is broken
           $ref: '#/components/responses/StatusResponse'
 
+
+  '/api/janitor/v1/resource/{type}':
```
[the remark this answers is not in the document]
wchamber: Yeah, we should just extract it, not have it in the parameter, instanceof is right.

src/main/resources/api/service_openapi.yaml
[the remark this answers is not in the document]
wchamber: Yeah, that's probably fine.

src/main/resources/api/service_openapi.yaml
```diff
@@ -79,6 +121,34 @@ components:
...
+        creationTime:
+          description: When the resource is created.
+          type: integer
+          format: int64
+
+    CreatedResource:
+      type: object
+      properties:
+        id:
+          type: string
```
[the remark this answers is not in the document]
wchamber: Seems reasonable to me.

.gitignore
wchamber: We should fix the gradle so that it can find the file. If we tell the gradle openApiGenerate rule that it depends on the resource files from `cloud-resource-schema`, we should be able to do do that rather than unzipping into src main. Not sure exactly what the gradle looks like, but let's try to get there. Let me know if you want to chat more about what this looks like.

src/main/resources/api/service_openapi.yaml
```diff
@@ -79,6 +111,34 @@ components:
...
+          type: object
+        labels:
+          description: The labels for the resource
+          type: object
+          additionalProperties:
+            type: string
+        timeToLiveInMinutes:
+          description: How long created resources should live (at least) before being cleaned up in minutes.
+          type: integer
+          format: int64
```
wchamber, line 129: int32 is probably fine

build.gradle
```diff
@@ -65,14 +70,27 @@ spotless {
...
+    cloudResourceSchema { transitive = false }
+}
+dependencies {
+    cloudResourceSchema group: 'bio.terra.cloud-resource-lib', name: 'cloud-resource-schema', version: '0.0.0-SNAPSHOT'
+}
+task unzipCloudResourceSchema(type: Copy) {
+    from zipTree(configurations.cloudResourceSchema.singleFile).matching {
+        include 'cloud_resources_uid.yaml'
+    }
+    into "$openapiTargetDir"
```
yonghaoy, line 91: I turned to just use the  openapiTargetDir instead puting it under builddir/src/main/resoruce/api

build.gradle
```diff
@@ -65,14 +70,27 @@ spotless {
...
 // - define dependencies to include the generated code
-
 def openapiSourceFile = "${projectDir}/src/main/resources/api/service_openapi.yaml"
 def openapiTargetDir = "${buildDir}/generated"
 
+configurations {
+    cloudResourceSchema { transitive = false }
+}
+dependencies {
+    cloudResourceSchema group: 'bio.terra.cloud-resource-lib', name: 'cloud-resource-schema', version: '0.0.0-SNAPSHOT'
```
wchamber, line 85: can we de-duplicate this (or at least the version) with what's on line 41?

yonghaoy: Maybe what we can do is 
(1) Declare the cloudResourceSchema configuration above. And merge those two dependencies.
(2) Let cloudResourceSchema's dependency externs implemeantion first
(3)Then we can have `cloudResourceSchema group: ...` thing once. 
But I feel that looks messy as we have things everywhere.  The current way is clean because we have one block for openAPI stuff and we can understand the steps dependencies. 
Feels like we didn't dedup anything and introducing extra work>

wchamber: version dedupe sgtm

PR #12, opened 2020-07-10
[CA 873] Create tracked resource endpoint

yonghaoy:


src/main/resources/api/service_openapi.yaml
```diff
@@ -138,6 +139,10 @@ components:
           type: object
           additionalProperties:
             type: string
+        creation:
+          description: The time when the resource is created.
+          type: string
+          format: 'date-time'
```
wchamber, line 145: did we change our minds about figuring this out in CRL with Instant.now()?

src/main/resources/api/swagger-config.json
```diff
@@ -7,6 +7,5 @@
   "generateApiTests": false,
   "generateApiDocumentation": false,
   "generateModelTests": false,
-  "generateModelDocumentation": false,
-  "library": "jersey2"
```
wchamber, line 11: did this not do something or are we switching to a different library? Maybe easier to be explicit about what library we're using.

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -50,18 +50,25 @@ databaseChangeLog:
               name: wait_until
               type: timestamp
               constraints:
-                nullable: false
+                nullable: true
                 unique: false
```
wchamber, line 54: I think we can remove this column as Stairway will handle waiting during a flight.

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -50,18 +50,25 @@ databaseChangeLog:
...
     - createTable:
         tableName: label
         columns:
+          - column:
+              name: id
+              type: uuid
+              constraints:
+                primaryKey: true
+                unique: true
+                nullable: false
```
wchamber, line 64: Why do we need a primary unique id for labels?

src/main/java/bio/terra/janitor/common/CloudResourceUidVisitor.java
```diff
@@ -0,0 +1,30 @@
+package bio.terra.janitor.common;
+
+import bio.terra.generated.model.*;
+
+public interface CloudResourceUidVisitor {
+  void visit(GoogleProjectUid resource);
```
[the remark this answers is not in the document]
yonghaoy: okay. I was concerned sub-visitor doesn't need return things, and I search in Google, someone mentioned the pattern could be having a getter method in sub-visitor and make the interface always `void`

src/main/java/bio/terra/janitor/common/JanitorResourceTypeEnum.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.janitor.common;
+
+/**
+ * Enums mapped from {@link null} to our DB schema. The original value can be found at <a
+ * href="https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml">terra-cloud-resource-lib
```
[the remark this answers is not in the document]
yonghaoy: Sorry, this is the old javadoc, forgot to update those. In my previous version this class did more things, it takes a cloudResourceUid and convert to janitor enum. Removed those things

src/main/java/bio/terra/janitor/common/exception/DuplicateLabelException.java
```diff
@@ -0,0 +1,12 @@
+package bio.terra.janitor.common.exception;
```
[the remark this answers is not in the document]
yonghaoy: This class are specific for janitor services, it only happens when you get key conflict when inserting data into our Label table. But the patten are the same(the supre calss - BadRequestException, ErrorReport exception). This is the pattern in WM and DB, for all errors they create different exceptions classes. If we (and I think) we use the same schema for error_report in response, many classes can be in common classes, and we extend them like this class. I had a TODO in GlobalExceptionHandler saying use common library once we have

src/main/java/bio/terra/janitor/common/exception/BadRequestException.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.janitor.common.exception;
+
+// This base class has data that corresponds to the ErrorReport model generated from
+// the OpenAPI yaml. The global exception handler auto-magically converts exceptions
+// of this base class into the appropriate ErrorReport REST response.
```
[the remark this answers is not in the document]
yonghaoy: Done. This class is copied from https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/common/exception/BadRequestException.java, didn't change anything...

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Autowired
+  public JanitorDao(JanitorJdbcConfiguration jdbcConfiguration) {
+    jdbcTemplate = new NamedParameterJdbcTemplate(jdbcConfiguration.getDataSource());
+  }
+
+  /** Creates the tracked_resource record and adding labels. */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public UUID createResource(
+      String cloudResourceUid,
```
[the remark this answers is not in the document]
yonghaoy: Need to convert this to json string before submit. And this dao will just do the insert operation without any data processing. And it assumes the parameters are already processed: date, json stuff process lives in service

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -0,0 +1,82 @@
...
+      JanitorResourceTypeEnum resourceType,
```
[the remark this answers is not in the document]
yonghaoy: Like my previous comment, I feel the janitorDao shouldn't do any data process work, it will just purely do the db logic - "insert" operation here. And data process depends on the upstream method who calls it.  So all information need to be provided explicitly when it reaches to the db layer, service layer handles converting request to db format, validations.

src/main/resources/db/changesets/20200625_initial_schema.yaml
yonghaoy: (1) A tracked resource will have multiple labels, and I am not sure a resource can have different values for the same key. 
(2) When handling duplicate labels, how we want to handle labels inside? Create new one or upgrade the existing ones?

If key + tracked_rource_id are enough, we can use those two as PK and remove this uuid?

src/main/resources/db/changesets/20200625_initial_schema.yaml
```diff
@@ -29,13 +29,13 @@ databaseChangeLog:
...
               name: creation
-              type: timestamp
+              type: timestamptz
               constraints:
                 nullable: false
                 unique: false
           - column:
               name: expiration
-              type: timestamp
+              type: timestamptz
```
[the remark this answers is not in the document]
yonghaoy: The openapi generated java code's timestamt is OffsetDateTime, and I see datarepo uses this:
https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/db/changesets/20181203_datasettable.yaml#L29, so kind of follow that pattern. I am working on some solution to convert OffsetDateTime to instant, if that works,  i can switch back to timestamp

src/main/resources/api/swagger-config.json
yonghaoy: I tries jersey2 and hope that can solve the oneOf problem, and seems it is not helpful. So no need to have this.
And I guess this wins:
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/build.gradle#L108

src/main/resources/db/changesets/20200625_initial_schema.yaml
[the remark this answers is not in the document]
wchamber: fwiw, I'm going to be a stickler about this one. Time zones and local times are a big mess. I don't think postgres or our api should refer to timezones, just "absolute" time. I spent a lot of time messing with timezones in my previous job and I have a healthy fear of them.

src/main/resources/db/changesets/20200625_initial_schema.yaml
wchamber: I think key + tracked_resource_id are enough for now. I could see us being in a world where we allow multiple values for the same key.
I don't think we're inserting new labels in many if all of our APIs. Let's handle the semantics of that when we get there.

src/main/java/bio/terra/janitor/common/exception/DuplicateLabelException.java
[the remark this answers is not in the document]
wchamber: Sorry, I was asking if these exceptions that are specific to the JanitorService/JanitorDao class should go in the packages that throw them instead of in the janitor.common.exception package. For example, data repo defines some exceptions for its JobService under that package even though those exceptions extend common exception types. https://github.com/DataBiosphere/jade-data-repo/tree/develop/src/main/java/bio/terra/service/job/exception

src/main/java/bio/terra/janitor/db/JanitorDao.java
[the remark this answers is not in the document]
wchamber: I disagree that CloudResourceUid serialization is not db specific logic. We logically want to store what's represented by the CloudResourceUid - we want to query based on that and get back answers that include CloudResourceUid. The JanitorDao is in charge of handling how to convert CloudResourceUid into a format that the database understands - which involves serializing it to a json string to use as jsonb postgres type.
Request parsing and validation are aspects of the service layer, so a `isValid(CloudResourceUid resource)` method is important to have for that layer, and the JanitorDao can state and assume it gets validated CloudResourceUids.

src/main/resources/db/changesets/20200625_initial_schema.yaml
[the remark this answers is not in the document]
wchamber: Sorry, looks like I'm not understanding postgres time types. This looks worrying, but here's what I found: https://jdbc.postgresql.org/documentation/head/8-date-time.html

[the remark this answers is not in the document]
wchamber: Ok, sorry. I guess we should use timestamptz in our database schema. In Java, let's have the JanitorDao work in Instants, and we can convert  to OffsetDateTime. before serializing to the database. Whatever we need to do in the openapi to make it think in Instants is also good, though I guess we can convert from OffsetDateTime if we have to.

src/main/resources/api/service_openapi.yaml
yonghaoy: ah, right. I was concerned client sets header randomly, and there are various date related field in header. But I think I l know how I can get this.

src/main/java/bio/terra/janitor/db/JanitorDao.java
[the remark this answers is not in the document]
yonghaoy: ah, okay, moved those to Dao. And when I didn't see if there is a need to add isValid() now. The logic will just do the serizie again? And I added a static helper classes, exception would be thrown there.

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -0,0 +1,82 @@
...
+      Map<String, String> labels,
+      OffsetDateTime creation,
+      OffsetDateTime expiration) {
```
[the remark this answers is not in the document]
yonghaoy: done

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -0,0 +1,82 @@
...
+        UUID labelId = UUID.randomUUID();
+        MapSqlParameterSource labelSqlParams =
+            new MapSqlParameterSource()
+                .addValue("id", labelId)
+                .addValue("tracked_resource_id", trackedResourceId)
+                .addValue("key", entry.getKey())
+                .addValue("value", entry.getValue())
+                .addValue("state", "READY");
+        try {
+          jdbcTemplate.update(insertLabelSql, labelSqlParams);
```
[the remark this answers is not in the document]
yonghaoy: done

src/main/resources/db/changesets/20200625_initial_schema.yaml
yonghaoy: okay! Also remove the state, because that should join the main table if we want that value.

src/main/java/bio/terra/janitor/common/exception/DuplicateLabelException.java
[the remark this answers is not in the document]
yonghaoy: I am concerned that those exception would be everwhere. I feel it is better to have a central package to all exceptions.
Not sure which way is better, WM did the janitor pattern I think.

src/main/java/bio/terra/janitor/common/JanitorResourceEnumVisitor.java
```diff
@@ -0,0 +1,48 @@
+package bio.terra.janitor.common;
+
+import static bio.terra.janitor.common.JanitorResourceTypeEnum.*;
+
+import bio.terra.generated.model.*;
+import bio.terra.janitor.common.exception.InvalidResourceUidException;
+import org.springframework.stereotype.Component;
+
+/** Gets {@link JanitorResourceTypeEnum} by visiting {@link CloudResourceUid}. */
+@Component
```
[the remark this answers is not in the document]
yonghaoy: Done

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    // Verify data are created correctly.
+    assertCreateResultMatch(
+        cloudResourceUid,
+        GOOGLE_PROJECT,
+        Optional.empty(),
+        Optional.empty(),
+        TIME_TO_LIVE_MINUTE,
+        jdbcTemplate,
+        DEFAULT_LABELS);
```
wchamber, line 62: I don't get why some of the TestUtils private static final variables are implicit in these functions, but only some of the time. e.g. TIME_TO_LIVE_MINUTE is inferred in newJsonCreateRequestBody, but is explicitly passed into assertCreateResultMatch. GOOGLE_PROJECT is used by newGoogleProjectResourceUid, but passed into assertCreateResultMatch. This makes it really hard to tell what's coming from where. I think the reason things are getting pulled into test helpers is because this is testing the same thing as the JanitorDao using the same methods. I think that's too much internal looking by this test.
Make things more explicit, or consistently implicit. Think if there's a test we can write that's testing less of the JanitorDao's implementation. Do less in TestUtils.

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
```diff
@@ -0,0 +1,51 @@
...
+    janitorDao.createResource(cloudResourceUid, DEFAULT_LABELS, CREATION, EXPIRATION);
+
+    assertCreateResultMatch(
+        cloudResourceUid,
+        GOOGLE_PROJECT,
+        Optional.of(CREATION),
+        Optional.of(EXPIRATION),
+        TIME_TO_LIVE_MINUTE,
+        jdbcTemplate,
+        DEFAULT_LABELS);
```
wchamber, line 49: ditto, odd that GOOGLE_PROJECT and TIME_TO_LIVE_MINUTE are implicit some of the time. I'd rather see this test be more explicit here, instead of relying on TestUtils to do everything.

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
yonghaoy: Done, per discussed offline, assert db record match will be the DaoTest's job. I spent quite a lot time trying to make a code reusable as much as possible.  Then the code looks confusion. Updated a bit, PTAL

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
yonghaoy: Refactor a little bit. Now in DaoTest, we will have assertCreateResultMatch in DaoTest, and it will call:
`assertCreateResultMatch(cloudResourceUid, GOOGLE_PROJECT, CREATION, EXPIRATION, DEFAULT_LABELS);`

I think the GOOGLE_PROJECT still make sense because the method above saying `CloudResourceUid cloudResourceUid = newGoogleProjectResourceUid();` so that is a google_project. Wdyt?

src/main/java/bio/terra/janitor/common/JanitorObjectMapperHelper.java
```diff
@@ -0,0 +1,17 @@
+package bio.terra.janitor.common;
+
+import bio.terra.generated.model.CloudResourceUid;
+import bio.terra.janitor.common.exception.InvalidResourceUidException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/** Helper function to wrap {@link ObjectMapper}. */
```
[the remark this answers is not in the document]
yonghaoy: not needed

src/main/java/bio/terra/janitor/common/exception/DuplicateLabelException.java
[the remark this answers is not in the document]
wchamber: sgtm. I asked Dan, about these and he said "I think the theory is that it is hard to predict over the long term which exceptions you will want to catch, so by using specific exceptions you keep your options open. But it is kind of like coding conventions where everybody has their own opinion and there isn’t one obvious right answer. I find it a little more helpful to have specific ones because the ide can help track them down more easily than a string search."
So my personal take would be to err towards less exceptions types until they're needed since we're not publishing this code as a library and can change it later without regard for backwards compatibility.

src/test/java/bio/terra/janitor/app/common/TestUtils.java
```diff
@@ -0,0 +1,107 @@
...
+    assertEquals(resourceType.toString(), actual.get("resource_type"));
+    creation.ifPresent(
+        time -> assertEquals(time, ((Timestamp) actual.get("creation")).toInstant()));
+    expiration.ifPresent(
+        time -> assertEquals(time, ((Timestamp) actual.get("expiration")).toInstant()));
+    assertEquals(
+        timeToLiveInMinutes,
+        Duration.between(
+                ((Timestamp) actual.get("creation")).toInstant(),
+                ((Timestamp) actual.get("expiration")).toInstant())
```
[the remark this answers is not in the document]
wchamber: yep.

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
wchamber: I think GOOGLE_PROJECT is still confusing. take it as an argument, or rename to DEFAULT_ and include that in the function. Or just inline the cloudResourceUid creation, it's not that long.

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -29,9 +30,13 @@ public JanitorDao(JanitorJdbcConfiguration jdbcConfiguration) {
...
 
-  /** Creates the tracked_resource record and adding labels. */
+  /**
+   * Creates the tracked_resource record and adding labels.
+   *
+   * <p>Note that we assume int input {@code cloudResourceUid} is valid.
+   */
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public UUID createResource(
+  public String createResource(
```
wchamber, line 39: I think this should return the TrackedResourceId. The fact that we have an id for the tracked resource is not specific to teh JanitorDao, e.g. we want to support gets with it and will do business logic per id. It's useful to have a type for it and not type it as a UUID or a String.

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -70,20 +75,31 @@ public UUID createResource(
...
+                          .addValue("tracked_resource_id", trackedResourceId.getUUID())
                           .addValue("key", entry.getKey())
                           .addValue("value", entry.getValue()))
               .toArray(size -> new MapSqlParameterSource[labels.size()]);
       try {
         jdbcTemplate.batchUpdate(insertLabelSql, sqlParameterSourceList);
       } catch (DuplicateKeyException e) {
-        throw new DuplicateLabelException(
+        throw new DuplicateDbKeyException(
             "Duplicate label found for tracked_resource_id: "
```
wchamber: I'm not sure this is worth catching and rethrowing specifically. But worrying about generating a conflicting UUID seems odd.
Thinking about this for the labels as well, with a Map<String, String> how can we have conflicting keys?

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -70,20 +75,31 @@ public UUID createResource(
...
       }
     }
-    return trackedResourceId;
+    return trackedResourceId.toString();
+  }
+
+  @VisibleForTesting
+  static String serialize(CloudResourceUid resource) {
+    ObjectMapper objectMapper =
+        new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_NULL);
```
wchamber, line 98: this needs a comment that it should not be changed since this is how the database will store the json and do serialization. I don't think postgres thinks that `{ "foo": null }` and `{}` are equivalent, even if we definitely want to store the second to save on space since they're logically equivalent in this case..

src/main/java/bio/terra/janitor/db/TrackedResourceId.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.janitor.db;
+
+import java.util.UUID;
+
+/** Wraps the tracked_resource_id in db tracked_resource table. */
+public class TrackedResourceId {
```
wchamber, line 6: adding a equals and hash method would be nice. Could just make this an autovalue.

src/main/java/bio/terra/janitor/db/exception/DuplicateDbKeyException.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.janitor.db.exception;
+
+import bio.terra.janitor.common.exception.BadRequestException;
+
+/**
+ * Exception when inserting into tracked_resource table while there is already a row with the same
```
wchamber, line 6: update comment, not just the tracked_resource table anymore

src/test/java/bio/terra/janitor/app/common/TestUtils.java
```diff
@@ -1,30 +1,15 @@
...
 import com.google.common.collect.ImmutableMap;
-import java.sql.Timestamp;
-import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.*;
-import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
-import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 public class TestUtils {
```
wchamber, line 12: Do we still need this file? I think it's reasonable to move the newJsonCreateREquestBody into the api controller test, since that's where' it's being used and it's reasonable to see the request shape in that test. I don't think we need to specify defaults values for the test that we have there.
The other static final fields can move into the only test that uses them, JanitorDaoTest.

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -41,29 +46,26 @@ public void setup() {
...
-        Optional.empty(),
-        TIME_TO_LIVE_MINUTE,
-        jdbcTemplate,
-        DEFAULT_LABELS);
+    assertResourceExists(response);
   }
 
   @Test
-  public void testCreateResourceFail_missingField() throws Exception {
+  public void createResourceFail_missingField() throws Exception {
```
wchamber, line 68: as-is, it's not very obvious what the missing field is. inlining the request creation or adding a comment would help.

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -72,4 +74,16 @@ public void testCreateResourceFail_missingField() throws Exception {
         .andDo(MockMvcResultHandlers.print())
         .andExpect(status().is4xxClientError());
   }
+
+  private void assertResourceExists(String jsonResponse) throws JsonProcessingException {
```
wchamber, line 78: maybe deserialize the response in the test proper and take a more meaningful type than String here. Links the known form of the json response more closely to where it's defined.

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
wchamber: sorry, I've been misreading it as a string, not as the ResourceType. Maybe don't fully import it so that we have `ResourceType.GOOGLE_PROJECT`.

yonghaoy: The resource type is not parameter `janitorDao.createResource`, the logic to get it happens inside the Dao, and it is done by the visitor. 
And the resource type is not part of cloudResourceUid creation. 
In the test here, I pass GOOGLE_PROJECT to assert method is because I know the CloudResourceUid is a google project(and it doesn't require GOOGLE_PROJECT as an input). How can I build the connection here between `newGoogleProjectResourceUid` with   `GOOGLE_PROJECT`? One way I can come up with is call visitor.visit(cloudResourceUid) to gee the type?

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
```diff
@@ -39,13 +50,62 @@ public void createTrackedResource() throws Exception {
     CloudResourceUid cloudResourceUid = newGoogleProjectResourceUid();
     janitorDao.createResource(cloudResourceUid, DEFAULT_LABELS, CREATION, EXPIRATION);
```
wchamber, line 51: should the created TrackedResourceId also be asserted against?

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
yonghaoy: ah I see what you mean, do you want me to put 
```
cloudResourceUid = CloudResourceUid()
        .googleProjectUid(new GoogleProjectUid().projectId(UUID.randomUUID().toString()));
```

In the test here, not to use the static method in TestUtils?

src/main/java/bio/terra/janitor/db/JanitorDao.java
yonghaoy: Done. 
I think we will need a class to represents most of the db entities.  Adding a class for TrackedResource was planed as part of the get endpoint ticket.

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
wchamber: yeah, though `projectId("my-project")` is probably clearer than using a uuid. Sorry about the GOOGLE_PROJECT, was confused by the enum.

src/main/java/bio/terra/janitor/db/JanitorDao.java
yonghaoy: Done. Removed this classes. I saw WM did that, so I did the same thing. But I now I am guessing the UUID is part of their request body, so it worth for them to catch it and throw a BadRequestException

src/main/java/bio/terra/janitor/db/JanitorDao.java
yonghaoy: added some comments. 
Actually I think Postgres works fine if we call to_jsonb(string). They are the same in json binary I guess?

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
yonghaoy: Did both

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
yonghaoy: Done

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: If that's true, let's not do NON_NULL and just write a more intrusive JanitorDaoTest that asserts that it's ok, like that we don't get back fields with null values in the raw json.

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
yonghaoy: I use the random UUID because each test run will insert one record into DB(DB is not getting reset) and I don't want those interrupt each other. 
Inlined the cloudResourceUid creation now..

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -75,24 +70,21 @@ public String createResource(
...
-            e);
-      }
+
+      jdbcTemplate.batchUpdate(insertLabelSql, sqlParameterSourceList);
     }
-    return trackedResourceId.toString();
+    return trackedResourceId;
   }
 
   @VisibleForTesting
```
wchamber, line 83: nit: move annotation below javadoc.

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: Still not resolved. Do we need NON_NULL for jsonb? Add a test if we don't.

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
```diff
@@ -0,0 +1,36 @@
...
+    return new CreatedResource()
+        .id(
+            janitorDao
+                .createResource(
+                    cloudResourceUid,
+                    body.getLabels(),
+                    creationTime,
+                    creationTime.plus(body.getTimeToLiveInMinutes(), ChronoUnit.MINUTES))
+                .id()
+                .toString());
```
wchamber, line 34: wdyt about making toString() on TrackedResourceId do this so we don't have to go through the UUID and we don't mess it up by accident?

src/main/java/bio/terra/janitor/db/TrackedResourceId.java
```diff
@@ -1,20 +1,22 @@
...
+  public static Builder builder() {
+    return new AutoValue_TrackedResourceId.Builder();
   }
 
-  public UUID getUUID() {
-    return this.id;
-  }
+  /** Builder for {@link TrackedResourceId}. */
+  @AutoValue.Builder
+  public abstract static class Builder {
```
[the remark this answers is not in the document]
yonghaoy: Actually I think it makse sense to generate in newBuilder() method? A UI should be auto generated when a newBuilder() getting creates?

src/main/java/bio/terra/janitor/db/exception/DuplicateDbKeyException.java
```diff
@@ -2,10 +2,7 @@
 
 import bio.terra.janitor.common.exception.BadRequestException;
 
-/**
- * Exception when inserting into tracked_resource table while there is already a row with the same
- * key.
- */
+/** Exception when inserting postgres table while there is already a row with the same key. */
 public class DuplicateDbKeyException extends BadRequestException {
```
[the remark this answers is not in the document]
yonghaoy: Like I asked in previous round of comments.  Do you think it make sense if we keep it as it may be used for future Dao's use cases? Or we should delete it for now and added it back if we need?

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -47,43 +53,63 @@ public void setup() {
...
-                    new ObjectMapper().readValue(jsonResponse, CreatedResource.class).getId()));
+        new MapSqlParameterSource().addValue("id", UUID.fromString(createdResource.getId()));
     assertEquals(1, jdbcTemplate.queryForMap(sql, params).size());
   }
+
+  private static String newJsonCreateRequestBody(
+      CloudResourceUid cloudResourceUid, Optional<Map<String, String>> labels) {
+    ObjectMapper mapper = new ObjectMapper();
+
+    ObjectNode trackedResourceNode =
```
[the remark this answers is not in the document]
yonghaoy: Done

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -47,43 +53,63 @@ public void setup() {
...
+    CloudResourceUid cloudResourceUid = new CloudResourceUid();
+
     this.mvc
         .perform(
             post("/api/janitor/v1/resource")
                 .contentType(MediaType.APPLICATION_JSON)
-                .content(newJsonCreateRequestBody(new CloudResourceUid(), Optional.empty())))
+                .content(newJsonCreateRequestBody(cloudResourceUid, Optional.empty())))
         .andDo(MockMvcResultHandlers.print())
         .andExpect(status().is4xxClientError());
```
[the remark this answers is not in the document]
yonghaoy: done

src/main/java/bio/terra/janitor/db/exception/DuplicateDbKeyException.java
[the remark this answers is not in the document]
yonghaoy: Deleted

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
yonghaoy: done

src/main/java/bio/terra/janitor/db/exception/DuplicateDbKeyException.java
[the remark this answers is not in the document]
wchamber: YAGNI. if we don't need it yet, don't add it yet.

src/main/java/bio/terra/janitor/db/TrackedResourceId.java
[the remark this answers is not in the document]
wchamber: No, I don't think TrackedResourceId should know that the way to create a new row is to generate a new UUID at random. That's up to the JanitorDao. TrackedResourceId is just a type holder for the UUID.

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: but keep some of the comment? It's still important to not mess with this for postgres.

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -100,16 +100,16 @@ private void assertResourceExists(CreatedResource createdResource) {
...
+        new CreateResourceRequestBody()
+            .resourceUid(cloudResourceUid)
+            .timeToLiveInMinutes(TIME_TO_LIVE_MINUTE);
     labels.ifPresent(
         l -> {
-          trackedResourceNode.set("labels", mapper.valueToTree(l));
+          body.setLabels(l);
         });
-    return trackedResourceNode.toString();
+    return new ObjectMapper().writeValueAsString(body);
```
wchamber: I don't think we need this method anymore. It's simple enough to create the CreateResourceRequestBody directly and serialize it.
Factory methods for objects that have a builder are an anti-pattern. If we want to have a DEFAULT_CREATE_REQUEST with some of the fields already initialized, that's a better testing pattern.

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
```diff
@@ -87,6 +89,23 @@ public void serializeCloudResourceUid() {
...
+   * Query by CloudResourceUid but not use the default serialize method which includes null value to
+   * verify query works.
+   */
+  public static Map<String, Object> queryTrackedResourceNonNull(
+      NamedParameterJdbcTemplate jdbcTemplate, CloudResourceUid resourceUid)
+      throws JsonProcessingException {
+    ObjectMapper mapper =
+        new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_NULL);
+    String sql =
+        "SELECT id, resource_uid::text, resource_type, creation, expiration, state FROM tracked_resource WHERE resource_uid::jsonb @> :resource_uid::jsonb";
```
wchamber, line 102: nit: divide long strings to keep them under the 100 column limit

wchamber, line 102: '@>' is not equality, its "contains top level fields." If this doesn't pass for '=' operator, then I don't think this works and we should use Include.NON_NULL so that we can add more resource types without messing up equality.
https://www.postgresql.org/docs/9.5/functions-json.html

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
yonghaoy: deleted

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
```diff
@@ -83,33 +81,14 @@ public void serializeCloudResourceUid() {
   public static Map<String, Object> queryTrackedResource(
       NamedParameterJdbcTemplate jdbcTemplate, CloudResourceUid resourceUid) {
     String sql =
-        "SELECT id, resource_uid::text, resource_type, creation, expiration, state FROM tracked_resource WHERE resource_uid::jsonb @> :resource_uid::jsonb";
+        "SELECT id, resource_uid::text, resource_type, creation, expiration, state " +
+                "FROM tracked_resource " +
+                "WHERE resource_uid::jsonb @> :resource_uid::jsonb";
```
wchamber, line 86: does this work with '=' instead of '@>'?

src/main/java/bio/terra/janitor/db/TrackedResourceId.java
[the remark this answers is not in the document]
wchamber: nit: I still think a factory function would be better than a builder.

src/test/java/bio/terra/janitor/db/JanitorDaoTest.java
yonghaoy: Yup!

PR #13, opened 2020-07-10
[CA 873] migrate from OpenApiGenerator to Swagger-Codegen for janitor service

yonghaoy:
Things has changed:

- Api Controller needs to override two new method getObjectMapper and getRequest which are not needed by us. So return empty for now.
- Not so many nulls -> in our case, there will be no null get setted if we don't put value in it(see UnauthenticatedApiControllerTest.java)
- Swagger codegen seems deleted the entire target dir first. So need to move the unziped cloud_resource_uid.yaml to another directory. 
- Minor syntax change in service_api yaml file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,6 @@ plugins {
     id 'idea'
     id 'com.google.cloud.tools.jib' version '2.4.0'
     id 'org.springframework.boot' version '2.3.1.RELEASE'
-    id 'org.openapi.generator' version '4.3.1'
     id 'io.spring.dependency-management' version '1.0.9.RELEASE'
     id 'com.diffplug.gradle.spotless' version '3.27.2'
     id 'org.hidetake.swagger.generator' version '2.18.2'
@@ -78,6 +77,7 @@ spotless {
 // - define dependencies to include the generated code
 def openapiSourceFile = "${projectDir}/src/main/resources/api/service_openapi.yaml"
 def openapiTargetDir = "${buildDir}/generated"
+def cloudResourceSchemaTargetDor = "${buildDir}/crlSchema"
 
 configurations {
     cloudResourceSchema { transitive = false }
@@ -89,7 +89,7 @@ task unzipCloudResourceSchema(type: Copy) {
     from zipTree(configurations.cloudResourceSchema.singleFile).matching {
         include 'cloud_resources_uid.yaml'
     }
-    into {buildDir}
+    into cloudResourceSchemaTargetDor
 }
 
 swaggerSources {
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ plugins {
     id 'org.openapi.generator' version '4.3.1'
     id 'io.spring.dependency-management' version '1.0.9.RELEASE'
     id 'com.diffplug.gradle.spotless' version '3.27.2'
+    id 'org.hidetake.swagger.generator' version '2.18.2'
 }
 
 group = 'bio.terra'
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -12,6 +12,7 @@ plugins {
     id 'org.openapi.generator' version '4.3.1'
```
wchamber, line 12: can we remove the openapi plugin?

build.gradle
```diff
@@ -91,42 +89,35 @@ task unzipCloudResourceSchema(type: Copy) {
     from zipTree(configurations.cloudResourceSchema.singleFile).matching {
         include 'cloud_resources_uid.yaml'
     }
-    into "$openapiTargetDir"
```
wchamber, line 94: should we remove the other referenes to 'openapi' in the comments and variables on lines 73-80? Or leave them, since we're still usin gthe openapi specification, just using swagger codegen to create the code? WDYT?

src/main/resources/api/swagger-config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "sortParamsByRequiredFlag": false,
```
wchamber, line 2: huh. Is there a eway to specify these fields in the gradle file? Ok if not, but putting these with the rest of the swagger codegen configuration in build.gradle seems preferable.

src/main/resources/api/service_openapi.yaml
```diff
@@ -121,8 +119,7 @@ components:
       properties:
         resourceUid:
           # CloudResourceUid is referred from CRL's cloud-resource-schema.
-          description: The UID of the resource in CRL format
-          $ref: '../../../../build/generated/cloud_resources_uid.yaml#/components/schemas/CloudResourceUid'
+          $ref: '../../../../build/cloud_resources_uid.yaml#/components/schemas/CloudResourceUid'
```
wchamber, line 122: putting cloud_resources_uid.yaml directly in the top level of `build` seems odd. Maybe put it in some subdirectory under `build` even if it can't live in the same directory as what's created by swagger code gen.

build.gradle
yonghaoy: I think OpenAPi is still the correct name. We just switch the the way to generate openapi java code from openApiGenator code to Swager-CodeGen

src/main/resources/api/swagger-config.json
yonghaoy: Look through all available options:
https://github.com/int128/gradle-swagger-generator-plugin#task-type-generateswaggercode seems configFIle is the only one works(configuration is something else).

And it only mentioned the json thing:
https://github.com/int128/gradle-swagger-generator-plugin#use-configuration-file

src/main/resources/api/service_openapi.yaml
yonghaoy: Created another dir for that. Thanks!

src/main/resources/api/service_openapi.yaml
```diff
@@ -34,8 +34,6 @@ paths:
       tags:
         - janitor
       requestBody:
-        description: |
-          Resource UID, time to live, resource labels
```
zarsky-broad, line 38: Does taking out the `|` and making this a non-literal string work? Or do you want to get rid of this?

yonghaoy: Even got complaints in the old line 124 where I don't have the `|`

zarsky-broad: Yeah, you can't have anything next to a $ref. You can use "allOf" to combine a ref and a description, though, or else just put the description inside the $ref. This line wouldn't have that problem, though.

PR #14, opened 2020-07-13
use jersey library which has a better oneof support

yonghaoy:
Seems jersery2 has a better support for oneof comepare with the default spring's one.


build.gradle
```diff
@@ -40,6 +40,7 @@ dependencies {
     implementation group: 'org.liquibase' , name: 'liquibase-core', version: '3.10.0'
     implementation 'org.springframework.boot:spring-boot-starter'
     implementation 'org.springframework:spring-jdbc:5.1.9.RELEASE'
+    implementation 'joda-time:joda-time:2.10.6'
```
wchamber, line 43: Uh, why do we need joda time? We should really stick to java8 time library. We don't want to mix&match with joda.

PR #15, opened 2020-07-14
fix few bugs in database schema

yonghaoy:
1. The test is not creating database so it is not catch errors in db schema
2. After turn on the database migration, find the error that ENUM type does not exist in liquidbase. 
3. Did the similar things as Stairway's solution, and Greg's previous comments "make enum as string in db while maintain it as Enum in code"
https://github.com/DataBiosphere/stairway/blob/a735a70d5045d33b376e400ab09e9f76e4bf4041/src/main/resources/stairway/db/changesets/20190118_base.yaml#L80

Also fix some readme typo...


README.md
```diff
@@ -43,7 +43,13 @@ You may need to use gcloud to provide GCR
   your target environment as the first argument:
 
 ```
-local-dev/setup_local_env.sh <environment>
+local-dev/setup_gke_deploy.sh <environment>
```
wchamber, line 46: should we update/remove local-dev/README?

PR #16, opened 2020-07-15
CA-941 Initial Stairway setup

wchamber:
Adds initial Stairway component to initialize Stairway on service start up.
Adds configuration to control Stairway.
Updates status endpoint to check Stairway.
Adds local sql changes to create a database and user for Stairway.
Modify unit test workflow to run on pushes to master and all pull requests, as well as uploading reports even if tests fail.
Add a shutdown endpoint for shutting down Stairway gracefully.


src/main/resources/api/service_openapi.yaml
```diff
@@ -25,6 +25,19 @@ paths:
           description: Service is broken
           $ref: '#/components/responses/StatusResponse'
 
+  # TODO configure this to only be callable by Kubernetes in production.
```
yonghaoy, line 28: why we need this endpoint?

src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
```diff
@@ -17,27 +18,48 @@
...
-              .ok(true)
-              .putSystemsItem("postgres", new SystemStatusSystems().ok(true)),
-          HttpStatus.OK);
+    SystemStatus systemStatus = new SystemStatus();
+
+    final boolean postgresOk =
+        jdbcTemplate.getJdbcTemplate().execute((Connection connection) -> connection.isValid(0));
+    systemStatus.putSystemsItem("postgres", new SystemStatusSystems().ok(postgresOk));
+
+    StairwayComponent.Status stairwayStatus = stairwayComponent.getStatus();
```
yonghaoy, line 38: I thought "subSystem" would be the "systems" janitor depends. And the one above already checks postgress. This one you added also checks postgres by using another db. And the thing you added also sounds like more than postgres healthy, it also includes "is db ready to use". Kind of this has overlaps with the previous one and we may need a clear definition about what level of scope the "subSystem" we want here.  An example in WM: https://github.com/DataBiosphere/terra-workspace-manager/blob/b19234a73a977898d0b77ac359f38631d71e5d3a/src/main/java/bio/terra/workspace/service/status/WorkspaceManagerStatusService.java

src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
```diff
@@ -17,27 +18,48 @@
...
-              .putSystemsItem("postgres", new SystemStatusSystems().ok(false)),
-          HttpStatus.INTERNAL_SERVER_ERROR);
+      return new ResponseEntity<>(systemStatus, HttpStatus.INTERNAL_SERVER_ERROR);
+    }
+  }
+
+  @Override
+  public ResponseEntity<Void> shutdownRequest() {
+    try {
+      if (!stairwayComponent.shutdown()) {
```
yonghaoy, line 55: are we shutting down everything or just stairway connection? That sounds this endpoint is partially done yet? Or this is just the endpoint  should be called "shutDownStariway"?

src/test/java/bio/terra/janitor/app/controller/UnauthenticatedApiControllerTest.java
```diff
@@ -25,13 +30,13 @@
...
-            .getResponse()
-            .getContentAsString());
+    MockHttpServletResponse response =
+        this.mvc.perform(get("/status")).andExpect(status().isOk()).andReturn().getResponse();
+
+    SystemStatus status =
+        new ObjectMapper().readValue(response.getContentAsString(), SystemStatus.class);
+    assertTrue(status.isOk());
+    assertThat(status.getSystems(), Matchers.hasKey("postgres"));
+    assertThat(status.getSystems(), Matchers.hasKey("stairway"));
```
yonghaoy, line 40: Would it be easier to call `/shutDown` then` /status` then we can verify both?

src/main/resources/application.properties
```diff
@@ -4,4 +4,12 @@ db.janitor.uri=jdbc:postgresql://127.0.0.1:5432/${DATABASE_NAME}
 db.janitor.username=${DATABASE_USER}
 db.janitor.password=${DATABASE_USER_PASSWORD}
 db.janitor.recreateDbOnStart=true
```
yonghaoy, line 6: unrelated question: I had `recreateDbOnStart = true` Because I was activaitly work on db things(also in GKE), and created bunch of dirty record, so I set this to true for now. Do you think we should keep as it or change it back to false?

wchamber: I think we'll need to change it back by the time we have persistent environments, but I don't think that's just yet.

src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
wchamber: Hmmm, yeah. I think at the end of the day the status should only return ok if the janitor service can operate, i.e. it's healthy. For stairway, that includes the database connection but also (soon, but not yet) a pubsub topic/subscription existing. I'd kind of like to leave those details of what Stairway needs to Stairway instead of introspecting it. I'm not sure how strictly to define a "subsystem." Being a little fuzzy about it seems ok to me.

src/main/resources/api/service_openapi.yaml
wchamber: Added a little more to the description of the endpoint, let me know if it's still not clear. Stairway in particular likes to shutdown gracefully so that less of its Flights are in a bad state.

src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
wchamber: The idea is to have k8s call this endpoint when it's going to terminate the pod (when it's being nice, there are no guarantees). This lets us leave around less garbage to recover later. Stairway has support for that, so we're calling Stairway. We might also stop scheduling our chronic tasks once we implement them.
Tried to add some comments, let me know if more would be helpful.

src/test/java/bio/terra/janitor/app/controller/UnauthenticatedApiControllerTest.java
wchamber: Not sure what you mean about ordering. Did add a test that helped find a bug in my application.properties, thanks.

yonghaoy: I mean since stairway already have a nice easy to shutdown things, then we can test the shutDown endpoint along with the status. i.e., add a test:
``` @Test
  public void tesShutDownAndServiceUnHealthy() throws Exception {       
     this.mvc.perform(get("/shutDown")).andExpect(status().isOk()).andReturn().getResponse();
    // Then get status endpoint and verify stairWay is not health..   
        this.mvc	
            .perform(get("/status"))	   
        SystemStatus status =
            .andExpect(status().**isNotOk**());```

src/main/resources/application.properties
yonghaoy: then do we need to do the similar things for the staritway db?

wchamber: Yes, eventually we need to change the values as well.
Note that Stairway#initialize does the recreate/migration, so we are calling that code already in this PR. https://github.com/DataBiosphere/stairway/blob/925d0724f57bc881e9f7bd1588bebd3818cc1732/src/main/java/bio/terra/stairway/Stairway.java#L351-L359

src/test/java/bio/terra/janitor/app/controller/UnauthenticatedApiControllerTest.java
wchamber: done.

PR #18, opened 2020-07-17
[CA 874] Add client library for janitor service

yonghaoy:
Would be great if we can have client library and easier for CRL
https://broadinstitute.jfrog.io/broadinstitute/webapp/#/artifacts/browse/tree/General/libs-snapshot-local/bio/terra/janitor/terra-janitor-client/0.0.0-SNAPSHOT/terra-janitor-client-0.0.0-20200717.152256-1.jar


.github/workflows/publish_client_library.yaml
```diff
@@ -0,0 +1,57 @@
+name: Publish package to Artifactory
+#on:
+#  release:
+#    types: [created]
+on:
+  push:
+    branches: [ yyu-CA-874 ]
```
wchamber, line 7: fix these before submitting. I'm also not sure we should do automatic library releases so much as manually trigger them. WDYT? Do workspace manager or jade-data-repo have an opinion?

.github/workflows/publish_client_library.yaml
```diff
@@ -0,0 +1,57 @@
...
+          echo ::add-mask::$ARTIFACTORY_USERNAME
+          echo ::set-output name=artifactory-password::$ARTIFACTORY_PASSWORD
+          echo ::add-mask::$ARTIFACTORY_PASSWORD
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Publish packages
+        run: ./gradlew artifactoryPublish
+        env:
+          ARTIFACTORY_USERNAME: ${{ steps.vault-secret-step.outputs.artifactory-username }}
+          ARTIFACTORY_PASSWORD: ${{ steps.vault-secret-step.outputs.artifactory-password }}
```
wchamber, line 57: nit: add a terminating newline

terra-resource-janitor-client/build.gradle
```diff
@@ -0,0 +1,96 @@
+plugins {
+    id 'java-library'
+    id 'maven-publish'
+
+    id 'com.jfrog.artifactory' version '4.16.0'
+    id 'org.hidetake.swagger.generator'
+}
+
+group = 'bio.terra.janitor'
+version = '0.0.0-SNAPSHOT'
```
wchamber, line 10: maybe put the version in settings.gradle

src/main/resources/api/service_openapi.yaml
```diff
@@ -138,6 +139,10 @@ components:
           type: object
           additionalProperties:
             type: string
+        creation:
+          description: The time when the resource is created.
+          type: string
+          format: 'date-time'
```
wchamber, line 145: ditto other PR, do we need this field? Maybe just omit it for now.

.github/workflows/publish_client_library.yaml
yonghaoy: I know WM do that manually. I had this because CRL did this. Feels like we shouldn't have this automattly, sounds release and publishing package are not the same thing. Let me remove this.

PR #20, opened 2020-07-22
[PF 22]: Add security schema for Janitor API

yonghaoy:
Part of https://broadworkbench.atlassian.net/browse/PF-22

### DIFF
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -9,9 +9,16 @@ info:
 servers:
   - url: /
 
+security:
+  - googleoauth:
+      - openid
+      - email
+      - profile
+
 paths:
   '/status':
     get:
+      security: []
       description: |
         Returns the operational status of the service
       operationId: serviceStatus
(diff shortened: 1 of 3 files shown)


terra-resource-janitor-client/build.gradle
```diff
@@ -20,6 +20,8 @@ dependencies {
     implementation group: 'org.glassfish.jersey.core', name: 'jersey-client', version: "${jersey}"
     implementation group: 'org.glassfish.jersey.media', name: 'jersey-media-json-jackson', version: "${jersey}"
     implementation group: 'org.glassfish.jersey.media', name: 'jersey-media-multipart', version: "${jersey}"
+    implementation "org.glassfish.jersey.inject:jersey-hk2:${jersey}"
```
wchamber, line 23: what do we need this for?

src/main/resources/api/service_openapi.yaml
```diff
@@ -186,3 +193,17 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/ResourceDescription'
+
+  securitySchemes:
```
wchamber, line 197: do we need to add `security[something]` in our yaml somewhere?

yonghaoy: Compared with SAM, Data Repo and WM code, seems they are similar
I think we won't need to add `security: []` for unthenticatevd APIs. For others, and seems we don't need that

terra-resource-janitor-client/build.gradle
yonghaoy: Ah, I think I can remove this one. I got some jersey binding exceptions in CRL when uses janitor calls. I saw data repo had that and it works. But I also added the same dependency in CRL, and it solves that problem. Let me delete this. 
Same issue as this: https://stackoverflow.com/questions/44088493/jersey-stopped-working-with-injectionmanagerfactory-not-found

PR #21, opened 2020-07-22
CA-955 Add more operations to JanitorDao to be used to schedule cleanup flights

wchamber:
- Add the flight table to the initial schema.
- Add value classes for representing records in the Janitor's database: TrackedResource and CleanupFlight.
- Use TrackedResource in create tracked resource endpoint and JanitorDao function.


src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -69,16 +69,110 @@ public TrackedResourceId createResource(
...
+        sql,
+        params,
+        (rs, rowNum) ->
+            TrackedResourceAndFlight.create(
+                TRACKED_RESOURCE_ROW_MAPPER.mapRow(rs, rowNum),
+                CLEANUP_FLIGHT_ROW_MAPPER.mapRow(rs, rowNum)));
+  }
+
+  /** Retrieve the {@link CleanupFlight}s associated with the tracked resource. */
+  public List<CleanupFlight> getFlights(TrackedResourceId trackedResourceId) {
```
yonghaoy, line 142: getFlightsByResourceId? Just in case we need other methods to query ClentupFlight by other things?

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -69,16 +69,110 @@ public TrackedResourceId createResource(
...
+        "SELECT flight_id, flight_state from cleanup_flight WHERE tracked_resource_id = :tracked_resource_id;";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource().addValue("tracked_resource_id", trackedResourceId.uuid());
+    return jdbcTemplate.query(sql, params, CLEANUP_FLIGHT_ROW_MAPPER);
+  }
+
+  /** Modifies the {@link CleanupFlightState} of a single flight. */
+  public void setFlightState(String flightId, CleanupFlightState flightState) {
+    String sql =
+        "UPDATE cleanup_flight SET flight_state = :flight_state " + "WHERE flight_id = :flight_id;";
```
yonghaoy, line 153: nit: new line or remove '+'

src/main/java/bio/terra/janitor/db/TrackedResource.java
```diff
@@ -0,0 +1,45 @@
+package bio.terra.janitor.db;
+
+import bio.terra.generated.model.CloudResourceUid;
+import com.google.auto.value.AutoValue;
+import java.time.Instant;
+
+/**
+ * A single tracked resource for persistence.
+ *
+ * <p>The Janitor service attempts to cleanup resources that it is tracking once they expire. This
```
yonghaoy, line 10: I am confused by this comments. What this tries to explain?  Does this need to be here?

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -69,16 +69,110 @@ public TrackedResourceId createResource(
...
-    return trackedResourceId;
   }
 
+  /**
+   * Retrieves and updates a TrackedResource that is ready and has expired by {@code now} to {@link
+   * TrackedResourceState#CLEANING}. Inserts a new {@link CleanupFlight} for that resource as
+   * initiating.
+   */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public Optional<TrackedResource> updateResourceForCleaning(Instant now, String flightId) {
```
yonghaoy, line 87: I didn't see how we use this. But here, do we want to do 1 by 1 or by `limit` at one time?  Seems currently it only process 1 record for each call, is the efficient?  And if the external method keep calling it, will they increase `now` for each call?

src/main/java/bio/terra/janitor/db/TrackedResource.java
```diff
@@ -0,0 +1,45 @@
...
+ * class describes a single resource being tracked.
+ */
+@AutoValue
+public abstract class TrackedResource {
```
yonghaoy, line 14: I would assume once we have the get endpoint, we would have a TrackedResource in openAPI. Seems WM and some DataRepo use that directly in DB. Then do we still need this? Or we just define it in openApi directly?

src/test/java/bio/terra/janitor/db/DatabaseTestUtils.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.janitor.db;
+
+import bio.terra.janitor.app.configuration.JanitorJdbcConfiguration;
+import bio.terra.janitor.service.migirate.MigrateService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/** Test utils for databases. */
+@Component
+public class DatabaseTestUtils {
```
yonghaoy, line 10: why we need this?  DB will get dropped and recreate by this:
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/test/resources/application.properties

wchamber:
> Question for how we represent db entities. I think JDBC has such support like @table annotation, id generation stuff:
> https://docs.spring.io/spring-data/jdbc/docs/current/reference/html/#jdbc.entity-persistence
> Do we want to use this?

That looks more like an ORM (object-relation mapping) system, which is controversial. Those veer pretty quickly into doing SQL for you, which many find probematic. I think Dan and Doug would both caution us to stay away from ORMs, but we can check.

src/test/java/bio/terra/janitor/db/DatabaseTestUtils.java
wchamber: That's between runs of the entire test suite, not per test.

src/main/java/bio/terra/janitor/db/TrackedResource.java
wchamber: I'd rather not mix layers of serialization. There may be things we want to add to the request API that we don't want in the database and vice versa. Locking different serialization layers has caused me pain in the past. It's easier to write thin conversion code to take `input request format` -> `internal representation` -> `database representation`.
The exception we have to that right now is CloudResourceUid, which I honestly don't feel great about, but it's supposed to be an external "standard" for us, so it's perhaps ok.

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: This is doing it 1 by 1. Each Stairway submission is going to be 1 by 1, so it's a little moot to worry about it. If we have performance issues, we can adjust.
It's up to the caller to adjust `now`. Getting the system time is going to be much smaller than network connection to the database. Or the caller can use a single `now` in repeated calls to just get things that had expired when they started and get the rest the next go around.

Yeah, these are not being used yet. Downside of breaking up PRs into smaller chunks is the whole picture isn't there at once. I can upload some WIP use for a look if you like.

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: I think `ByResourceId` encoded in the input parameters. I'm ok overloading this method, or we can differentiate later when we add more methods.

src/main/java/bio/terra/janitor/db/TrackedResource.java
wchamber: This is trying to document/explain what a tracked resource is. It's kind of fundamental to what the Janitor is doing, but we should still describe it somewhere. Having it on the value class that we're going to use everywhere to represent a "tracked resource" seemed like a good idea.
Very open to help here. Any ideas about how to make this better.

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: done.

src/main/java/bio/terra/janitor/db/JanitorDao.java
yonghaoy: (1) 1 by 1 make sense then if that is what stariway needs
(2) If the `now` is the adjustable, wdyt not to use now term, and this method would be updateRecordForCleaning at certain timestamp, and the timestamp can be any things depends what upstream system needs?
(3) WIP pr is not needed. We can always adjust the code here in follow PRs where it is used.

src/test/java/bio/terra/janitor/db/DatabaseTestUtils.java
yonghaoy: Ah, we have the same pain, then I end up using random UUID as the project name instead of a fixed value. 
Actually the DB is easily get messed up, there should be some way to tell Spring reset per test run. 
one think I just find out is to use this:
`@DirtiesContext(classMode = ClassMode.AFTER_EACH_TEST_METHOD)` But I haven't try it yet.
https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/test/annotation/DirtiesContext.html
Do you want to try this out?

src/main/java/bio/terra/janitor/db/TrackedResource.java
yonghaoy: How about adding something like 
"this class mapped to tracked_resource table in db". And what is tracked_resource table is explained in db schema?

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: 2) Open to different parameter names. I think `now` is reasonable because it captures the intent - what is ready to clean depends on what time it is "now" as determined by the caller. You should be calling this with a notion of "now," not next week. Brainstorming other parameters names: `expirationAtLeast`, `expiredBy`, other ideas?

src/test/java/bio/terra/janitor/db/DatabaseTestUtils.java
wchamber: Thanks that works.
We might get to the point where it's painfully slow to recreate the entire app each time, but let's deal with that if/when we get there.

src/main/java/bio/terra/janitor/db/TrackedResource.java
wchamber: tweaked, ptal.

src/test/java/bio/terra/janitor/db/DatabaseTestUtils.java
yonghaoy: ah, okay, drop and creating db also take some time, but I agree restart server would take bit longer than that.

src/main/java/bio/terra/janitor/db/JanitorDao.java
yonghaoy: expiredBy or the other SG to me. 
I think the logic for this Dao's method doesn't tight to ''now;' context. Just from reading the code and understanding the logic, seems it just updateRecordForCleaning by the time and flight id whatever upstream layer gives to it.

wchamber: `expiredBy` done.

PR #22, opened 2020-07-23
PF-27 Add FlightManager class for interacting with the Janitor DB and Stairway

wchamber:


src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -0,0 +1,113 @@
...
+    int submissions = 0;
+    for (JanitorDao.TrackedResourceAndFlight resourceAndFlight : resourceAndFlights) {
+      String flightId = resourceAndFlight.cleanupFlight().flightId();
+      try {
+        // If there is some flight state, then the flight has been submitted successfully and does
+        // not need to be recovered. There's nothing to do but wait for the flight to update it's
+        // state from INITIATING.
+        // TODO(wchamber): add metrics.
+        stairway.getFlightState(flightId);
+      } catch (FlightNotFoundException e) {
```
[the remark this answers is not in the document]
wchamber: Yeah, it's not great. I think this is the only way that Stairway supports at the moment. We can make a feature request.
https://github.com/DataBiosphere/jade-data-repo/blob/c29b436e61672f4517adb45a3630efa8d1a1245c/src/main/java/bio/terra/service/job/JobService.java#L515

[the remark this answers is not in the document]
wchamber: This is kind of expected behavior, I'm not sure what information is worth logging. If we did log something, it would be at info. I think the caller of this method may log how many flights were recovered total, but logging each flight isn't particularly interesting I think.

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -0,0 +1,113 @@
...
+   * Stairway. Resubmit the flights, returning how many flights were resubmitted.
+   *
+   * <p>This function assumes that it is not running concurrently with any other submissions to
+   * Stairway, e.g {@link #submitFlight(Instant)}.
+   */
+  public int recoverUnsubmittedFlights() {
+    List<JanitorDao.TrackedResourceAndFlight> resourceAndFlights =
+        janitorDao.retrieveResourcesWith(CleanupFlightState.INITIATING, RECOVERY_LIMIT);
+    if (resourceAndFlights.size() == RECOVERY_LIMIT) {
+      logger.error(
```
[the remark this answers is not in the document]
wchamber: If we're even close to the limit, it's going to be a problem soon. We'd really like to recover everything at startup and if we're right at the limit, we need to address something.
I could see us making this a warning, but we really would want to address this if it happens. I need to add a gauge metric for us to alert on, but going to do that later.

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /** Submits a cleanup flight for the resource to Stairway, or fails logging any exceptions. */
+  private void submitToStairway(String flightId, TrackedResource resource) {
+    FlightFactory.FlightSubmission flightSubmission =
+        cleanupFlightFactory.createSubmission(resource);
+    try {
+      stairway.submitToQueue(
+          flightId, flightSubmission.clazz(), flightSubmission.inputParameters());
+    } catch (StairwayException | InterruptedException e) {
+      logger.error(
```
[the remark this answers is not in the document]
wchamber: good catch, done.

src/main/java/bio/terra/janitor/service/cleanup/FlightFactory.java
```diff
@@ -0,0 +1,27 @@
...
+import bio.terra.janitor.db.TrackedResource;
+import bio.terra.stairway.Flight;
+import bio.terra.stairway.FlightMap;
+import com.google.auto.value.AutoValue;
+
+/**
+ * An interface for creating the parameters needed to submit a {@link Flight} to Stairway to cleanup
+ * a {@link TrackedResource}.
+ */
+public interface FlightFactory {
```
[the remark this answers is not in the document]
wchamber: done

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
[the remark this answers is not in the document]
yonghaoy: alert sounds a good idea? Can we add a TODO here just in case we forgot?

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /** Submits a cleanup flight for the resource to Stairway, or fails logging any exceptions. */
+  private void submitToStairway(String flightId, TrackedResource resource) {
+    FlightFactory.FlightSubmission flightSubmission =
+        cleanupFlightFactory.createSubmission(resource);
+    try {
+      stairway.submitToQueue(
+          flightId, flightSubmission.clazz(), flightSubmission.inputParameters());
+    } catch (StairwayException | InterruptedException e) {
```
yonghaoy, line 104: Seems you just catch the exception and log it. Then you still increase the submissions by one in the caller method. Do we want to return boolean success, and don't do the increase if `success==false?`

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -0,0 +1,113 @@
...
+  private Logger logger = LoggerFactory.getLogger(FlightManager.class);
+
+  /**
+   * A limit on how many flights to recover at startup. This is to prevent blowing out the service
+   * at start up, but if we have too many flights to recover, they will not all be recovered
+   * immediately.
+   */
+  private static final int RECOVERY_LIMIT = 1000;
+
+  private final Stairway stairway;
```
yonghaoy, line 35: Can we use dependency injection for this class? 
i.e., autowire the three parameters and make this component?

src/test/java/bio/terra/janitor/service/cleanup/FlightManagerTest.java
```diff
@@ -0,0 +1,182 @@
...
+            stairwayComponent.get(),
+            janitorDao,
+            trackedResource ->
+                FlightFactory.FlightSubmission.create(OkCleanupFlight.class, new FlightMap()));
+    assertFalse(manager.submitFlight(EXPIRATION).isPresent());
+  }
+
+  @Test
+  public void recoverUnsubmittedFlights_unsubmittedFlight() throws Exception {
+    FlightManager manager =
```
yonghaoy, line 111: ditto, we should make injectable

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
wchamber: We don't have a FlightSubmissionFactory to wire for this class yet. And I don't actually want this class to be available from anywhere in the applciation, so not going to autowire this class itself.

src/test/java/bio/terra/janitor/service/cleanup/FlightManagerTest.java
wchamber: ditto, don't want to expose FlightManager to the entire application.

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
wchamber: done

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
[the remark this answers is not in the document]
wchamber: done

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
yonghaoy: (1) Just to confirm what I understand: For now we don't have FlightSubmissionFactory to wire so we are not injecting 
staritway, janitorDaom, submissionFactory yet. Instead, we will manually creates/pass them from upsteam services? 
But once we have FlightSubmissionFactory to wire, we will switch to inject them,
(2) We will not make the class a compoment? 
(3) Also, make the class injectable vs controling the usage of the class are different things. We can always have some ways to control the access

wchamber: 1) Yes. See https://github.com/DataBiosphere/terra-resource-janitor/pull/24
2) I don't think we need to, so I don't know why we would.

PR #23, opened 2020-07-24
Upgrade to released version of stairway:0.0.20.

wchamber:
Add try-catch to stairway build since it now throws a checked exception.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,9 @@ repositories {
     maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/'
     }
+    maven {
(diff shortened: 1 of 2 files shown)


wchamber:
@ddietterich fyi

build.gradle
```diff
@@ -23,6 +23,9 @@ repositories {
     maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/'
     }
+    maven {
+        url 'https://broadinstitute.jfrog.io/broadinstitute/libs-release-local/'
```
yonghaoy, line 27: can we move this into the `maven` block above?

wchamber: don't think so. I think each of these is a different configuration of a maven repository - can't configure multiple repositories at once.

PR #24, opened 2020-07-24
PF-27 Create scheduler to periodically run the FlightMonitor to schedule flights.

wchamber:
- Add a FlightSubmissionFactory that makes flights that fail to clean up all types of resources. Actual cleanup impls coming later.


src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
```diff
@@ -53,6 +58,7 @@
   @Override
   public ResponseEntity<Void> shutdownRequest() {
     try {
+      flightScheduler.shutdown();
```
yonghaoy, line 61: One unrelated thing I notice: The current shutdown is unauthenticatated api.
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/resources/api/service_openapi.yaml#L43
Can you also update it here, I don't think we want anybody to shut this down..
But you said this is only callable by k8s, that means we have to put it unauthenticatated to bypass the apache proxy? And after that, I am not sure how we do authZ. (would be easier once we know what the request header looks like for calls from k8s)

src/main/java/bio/terra/janitor/service/cleanup/FlightScheduler.java
```diff
@@ -0,0 +1,90 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/** The FlightScheduler runs the {@link FlightManager} periodically to clean resources. */
+// TODO add metrics.
+@Component
+public class FlightScheduler {
+  /** How often to query for flights to schedule. */
+  private static final Duration SCHEDULE_PERIOD = Duration.ofMinutes(1);
```
yonghaoy, line 22: Just curious how we get this number? Seems the schedule runs every one minute and do the `scheduleFlights`. And the `scheduleFlights` will keep do things until empty.  Will scheduleFlights usually takes longer than 1 minute to finish and we just want the schedule things last forever with no delay? Or we want to give it a bigger number than the normal one `scheduleFlights` run?

src/test/java/bio/terra/janitor/service/cleanup/FlightSchedulerTest.java
```diff
@@ -0,0 +1,78 @@
...
+  @Autowired PrimaryConfiguration primaryConfiguration;
+  @Autowired JanitorDao janitorDao;
+
+  @Test
+  public void resourceScheduledForCleanup() throws Exception {
+    Instant now = Instant.now();
+    TrackedResource resource =
+        TrackedResource.builder()
+            .trackedResourceId(TrackedResourceId.create(UUID.randomUUID()))
+            .trackedResourceState(TrackedResourceState.READY)
```
yonghaoy, line 42: I thought when we create resource, we don't need to pass this value, this is done inside dao's `createResource` method

src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
wchamber: I'm not sure how we'll do this either, but we should be able to use whatever datarepo is doing for their shutdown endpoint. Filed PF-57 and TODO'd it here. Data repo also has their endpoint as unauthenticated.
Ok to leave any changes here until that bug? We should do it before having "production" deployments. We could pick it up sooner if it feels important.

src/main/java/bio/terra/janitor/service/cleanup/FlightScheduler.java
wchamber: I do not know what the right numbers are here, and they'll probably change with different workloads. No idea yet how long this will typically take. I don't think we want to look for new resources more than once a minute, so that seemed like an ok place to start.
If scheduleFlights does not complete in less than a minute, a new scheduleFlights will not be started by the ScheduledExecutorService until the previous call completes. So we don't have to worry about them running simultaneously. Adding a comment.

src/test/java/bio/terra/janitor/service/cleanup/FlightSchedulerTest.java
wchamber: Not since PR#21. I'm moving towards making the JanitorDao dumber and just do the operations of retrieving/modifying the persistent state. That resources often start in the READY state is business logic. (Note that they may not start in the READY 
state if they duplicate another resource. I don't think we want all of the dedupe logic to live in the JanitorDao). Along those lines, I think I'm going to pull the logic out of JanitorDao#updateResourceForCleaning into the FlightMonitor in a follow up PR.

PR #26, opened 2020-07-27
PF-26 Add Flight management for completed flights.

wchamber:


src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -139,31 +169,37 @@ public void createResource(TrackedResource resource, Map<String, String> labels)
...
     String sql =
         "SELECT flight_id, flight_state from cleanup_flight WHERE tracked_resource_id = :tracked_resource_id;";
     MapSqlParameterSource params =
         new MapSqlParameterSource().addValue("tracked_resource_id", trackedResourceId.uuid());
     return jdbcTemplate.query(sql, params, CLEANUP_FLIGHT_ROW_MAPPER);
   }
 
-  public Optional<CleanupFlightState> getFlightState(String flightId) {
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public Optional<CleanupFlightState> retrieveFlightState(String flightId) {
```
yonghaoy, line 182: just curious: Why we want to rename get to retrive for all of them?

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -139,31 +169,37 @@ public void createResource(TrackedResource resource, Map<String, String> labels)
...
   }
 
-  /** Modifies the {@link CleanupFlightState} of a single flight. */
-  public void setFlightState(String flightId, CleanupFlightState flightState) {
+  /**
+   * Modifies the {@link CleanupFlightState} of a single flight. Returns whether there was a flight
+   * with the flightId.
+   */
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public boolean updateFlightState(String flightId, CleanupFlightState flightState) {
```
yonghaoy, line 195: the updateTrackedResource method you returned Optional<TrackedResource>, but here, you only returns boolean. Do we want to have some consistent guideline about the return for an update operation?

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -118,5 +120,95 @@ private boolean submitToStairway(String flightId, TrackedResource resource) {
     return true;
   }
 
-  // TODO(wchamber): Add methods for finding completed and fatal Flights.
+  /**
+   * Find up to {@code limit} flights that are finishing cleanup in the Janitor's storage and
```
yonghaoy, line 124: finds

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -118,5 +120,95 @@ private boolean submitToStairway(String flightId, TrackedResource resource) {
...
+      // The resource should not have moved from CLEANING to any other state while there was a
+      // flight working on it.
+      throw new UnexpectedCleanupState(
+          String.format("Unexpected TrackedResourceState: %s", resourceState));
+    }
+    janitorDao.updateFlightState(flightId, CleanupFlightState.FINISHED);
+  }
+
+  /** Exception for unexpected resource state when finishing a cleanup flight. */
+  private static class UnexpectedCleanupState extends Exception {
```
yonghaoy, line 207: UnexpectedCleanupStateException? And do we need to make this has its own class?

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -118,5 +120,95 @@ private boolean submitToStairway(String flightId, TrackedResource resource) {
...
+    if (!resource.isPresent()) {
+      throw new UnexpectedCleanupState(
+          String.format("Unable to find tracked_resource with id [%s]", trackedResourceId));
+    }
+    TrackedResourceState resourceState = resource.get().trackedResourceState();
+    if (resourceState.equals(TrackedResourceState.CLEANING)) {
+      janitorDao.updateResourceState(trackedResourceId, endCleaningState);
+      // We assume no one else is modifying the CleanupFlightState while we do this.
+    } else if (!resourceState.equals(TrackedResourceState.ABANDONED)
+        && !resourceState.equals(TrackedResourceState.DUPLICATED)) {
```
yonghaoy, line 197: Having duplicates state is expected. Feel this should not be an exception.

src/main/java/bio/terra/janitor/service/cleanup/flight/UnimplementedCleanupStep.java
```diff
@@ -0,0 +1,21 @@
+package bio.terra.janitor.service.cleanup.flight;
+
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+
+/** A fatal step for resource types that have not been implemented yet. */
+public class UnimplementedCleanupStep implements Step {
```
yonghaoy, line 9: nit: Unsupported?

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: I was thinking that `get` might imply it's like a getter where the info is in memory, whereas `retrieve` sounds more like a network operation. And I think I was already being inconsistent, so moved to all `retrieve`. sg?

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: There's `retrieveTrackedResource` that doesn't modify and returns an Optional.
There's `updateFlightState`that does modify and returns a boolean.
There's `updateResourceForCleaning` that does modify and returns an Optional. I'm going to refactor it to not have the logic and not do an update that returns the optional, so yes.
Does that sg?

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
wchamber: done

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
wchamber: So we need to throw something in order to get the rollback. I thought it made sense for it to be its own class so that we can catch only violations to our own logical invariants and log them accordingly. Then we can handle them separately from unexpected exceptions. But those are both forms of unexpected, so we could not special case them and throw something like `IllegalStateException`. WDYT?

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
wchamber: Not sure what you mean, there's a `!` negation. Anything that's *not* ABANDONED and DUPLICATED is not expected. DUPLICATED is expected and does not throw.

src/main/java/bio/terra/janitor/service/cleanup/flight/UnimplementedCleanupStep.java
wchamber: done.

PR #28, opened 2020-07-28
[PF-9] Add bearer token into request's auth header

yonghaoy:
https://broadworkbench.atlassian.net/browse/PF-9

### DIFF
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -10,10 +10,8 @@ servers:
   - url: /
 
 security:
-  - googleoauth:
-      - openid
-      - email
-      - profile
+  - bearerAuth: []
+  - authorization: [openid, email, profile]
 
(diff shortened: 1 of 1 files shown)


yonghaoy:
This is how WM did that. 
https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/resources/api/service_openapi.yaml#L595
`Having bearerAuth: [] `in security will let the endpoint to use the bearAuth defined in our `SecuritySchema` below(TDR and SAM is using the 2.0's `SecurityDefination`). Without `bearerAuth: []`, our endpoint won't have the auth header. 
That is what I find out before last Friday's demo. Having that field, our Apache Proxy(same as WM's one in terra-helm) is able to grab token from the bearAuth field and do the authN. TDR and SAM is using differernt apache template, and I tried theirs, seems there is no auth field in their header as well.

PR #30, opened 2020-07-29
PF-23 Add management and scheduling for fatal flights.

wchamber:


src/main/java/bio/terra/janitor/app/configuration/PrimaryConfiguration.java
```diff
@@ -30,6 +30,12 @@
   /** How many flights to process for completion at a time. */
   private int flightCompletionLimit = 1000;
 
+  /** How often to query for flights that have completed fatally. */
+  private Duration fatalFlightCompletionPeriod = Duration.ofMinutes(1);
```
yonghaoy, line 34: Do we need different configs for the differrent state? They looks the same now. Is there any known reason we may have different period and limit?

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -188,18 +201,38 @@ public boolean updateResourceState(
...
         new MapSqlParameterSource()
             .addValue("flight_state", flightState.toString())
             .addValue("flight_id", flightId);
-    return jdbcTemplate.update(sql, params) == 1;
+    return Optional.ofNullable(
+        DataAccessUtils.singleResult(jdbcTemplate.query(sql, params, CLEANUP_FLIGHT_ROW_MAPPER)));
+  }
+
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public Map<String, String> retrieveLabels(TrackedResourceId trackedResourceId) {
```
yonghaoy, line 222: Is this needed by flight or just something we will need in the future?

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -188,18 +201,38 @@ public boolean updateResourceState(
...
+        new MapSqlParameterSource().addValue("tracked_resource_id", trackedResourceId.uuid());
+    List<AbstractMap.SimpleEntry<String, String>> labels =
+        jdbcTemplate.query(
+            sql,
+            params,
+            (rs, rowNum) ->
+                new AbstractMap.SimpleEntry<>(rs.getString("key"), rs.getString("value")));
+    return labels.stream()
+        .collect(
+            Collectors.toMap(AbstractMap.SimpleEntry::getKey, AbstractMap.SimpleEntry::getValue));
```
yonghaoy, line 235: The tracked UID + key is the PK. There is no validations to check the duplicates of a key(not sure if we will have duplicate keys for a resource). If we want this way, we may need to add some checks during resource creation?

src/main/java/bio/terra/janitor/app/configuration/PrimaryConfiguration.java
wchamber: Yeah, they could process at different speeds, so we might want to give them shorter or longer periods. We might also see different rates for the limits.
Started them the same since I don't know which way they'll go.

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: Oops, this was already included in the previous PR #29, just hadn't done the right merging yet.

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: That's already enforced by the JanitorDao API. Since we only can create Labels using a Map<String, String> with the tracked resource id, there's not a way to violate that. The sql insertion would also fail wby violating the foreign key constraint, so we don't need to check that postgres is maintaining its variants on retrieval.

PR #32, opened 2020-07-31
[PF 61] Janitor track resource using pubsub. Also setup integration test environment

yonghaoy:
For now, I added a retrieveTrackedResource(CloudResourceUid uid) to verify resource are created. Will switch to the get endpoint once we have that.


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,95 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Buind And Run Tests
```
wchamber, line 4: buind typo

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,95 @@
+# This workflow will build a Java project with Gradle
```
wchamber, line 1: build_and_test seems a little redundant. maybe just test.yml? (I know that's what we have on CRL atm, but doesn't seem like a great filename).
also, it would be nice if git tracked the history of this file from unit_test.yml, if that's not too much work. not necessary.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,95 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ '**' ]
+env:
+  VAULT_ADDR: https://clotho.broadinstitute.org:8200
+  VAULT_SERVICE_ACCOUNT_PATH: secret/dsde/terra/kernel/integration/yyu/crl_janitor/app-sa
+  VAULT_CLIENT_SERVICE_ACCOUNT_PATH: secret/dsde/terra/kernel/integration/yyu/crl_janitor/client-sa
```
wchamber, line 14: Is the plan to use the toolsalpha path instead? TODO?

build.gradle
```diff
@@ -50,13 +50,17 @@ dependencies {
     implementation 'bio.terra:stairway:0.0.20'
     implementation group: 'bio.terra.cloud-resource-lib', name: 'cloud-resource-schema', version: crlSchemaVersion
 
+    // implementation 'com.google.cloud:google-cloud-pubsub:1.104.1'
+
+    compile group: 'com.google.cloud', name: 'google-cloud-pubsub', version: '1.104.1'
```
wchamber, line 55: why is it compile and not implementation? remove commented out line regardless.

src/main/java/bio/terra/janitor/app/configuration/TrackResourcePubsubConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.stereotype.Component;
+
+/** Pub/sub Configurations for tracking resource. */
+@Component
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "pubsub.track-resource")
```
wchamber, line 12: let's be consistent and continue to camelCase, so `pubsub.trackResource`

src/main/java/bio/terra/janitor/app/configuration/TrackResourcePubsubConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+public class TrackResourcePubsubConfiguration {
+  private boolean pubsubEnabled;
```
wchamber, line 14: nit: maybe just `enabled`? I don't think there's any part that's not enabled if this field is false.

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -88,6 +87,18 @@ public void createResource(TrackedResource resource, Map<String, String> labels)
...
+  /** Returns the {@link TrackedResource} for a {@link CloudResourceUid} if there is one. */
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public Optional<TrackedResource> retrieveTrackedResource(CloudResourceUid cloudResourceUid) {
+    String sql =
+        "SELECT id, resource_uid, creation, expiration, state FROM tracked_resource tr "
+            + "WHERE resource_uid::jsonb = :resource_uid::jsonb";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource().addValue("resource_uid", serialize(cloudResourceUid));
+    return Optional.ofNullable(
+        DataAccessUtils.singleResult(jdbcTemplate.query(sql, params, TRACKED_RESOURCE_ROW_MAPPER)));
```
wchamber, line 99: this is ok for this PR, if you add a TODO to remove it later.
We don't want this long term because there's no database guarantee that there will be at most one result for this query.

src/main/java/bio/terra/janitor/service/pubsub/TrackedResourceSubscriber.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.cloud.pubsub.v1.AckReplyConsumer;
+import com.google.cloud.pubsub.v1.MessageReceiver;
+import com.google.cloud.pubsub.v1.Subscriber;
+import com.google.pubsub.v1.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/** Subscribes track resource message and update Janitor record. */
```
wchamber, line 16: nit: `Subscribes to tracked resource topic to update the Janitor database for new tracked resources.`

src/main/java/bio/terra/janitor/service/pubsub/TrackedResourceSubscriber.java
```diff
@@ -0,0 +1,65 @@
...
+  TrackedResourceSubscriber(
+      TrackResourcePubsubConfiguration trackResourcePubsubConfiguration,
+      JanitorService janitorService) {
+    this.trackResourcePubsubConfiguration = trackResourcePubsubConfiguration;
+    this.janitorService = janitorService;
+  }
+
+  public void initialize() {
+    if (!trackResourcePubsubConfiguration.isPubsubEnabled()) {
+      return;
```
wchamber, line 32: let's add a log info for whether or not this is initialized. Helps for sanity checking that a service has started up as expected.

src/main/java/bio/terra/janitor/service/pubsub/TrackedResourceSubscriber.java
```diff
@@ -0,0 +1,65 @@
...
+          try {
+            CreateResourceRequestBody body =
+                new ObjectMapper()
+                    .readValue(message.getData().toStringUtf8(), CreateResourceRequestBody.class);
+            janitorService.createResource(body);
+          } catch (JsonProcessingException e) {
+            throw new InvalidMessageException(
+                "Invalid track resource pubsub message: " + message.toString(), e);
+          } finally {
+            consumer.ack();
```
wchamber, line 57: I don't think we should always ack the message. We should only ack it if we've persisted correctly. If we throw a database exception, the message should not be acked.

I think we should set up a dead letter queue for messages we're not able to process of silently dropping them. Then we can add monitoring/alerting and see what ends up there. https://cloud.google.com/pubsub/docs/dead-letter-topics
Even invalid format message should not be dropped silently - what if there's a bug in the janitor's processing of the request?
Leave a TODO for now instead of implementing the DLQ to not make this PR too big, but only ack if we're successful.

src/main/java/bio/terra/janitor/service/pubsub/TrackedResourceSubscriber.java
```diff
@@ -0,0 +1,65 @@
...
+          }
+        };
+
+    Subscriber subscriber = Subscriber.newBuilder(subscriptionName, receiver).build();
+    // Start the subscriber.
```
wchamber, line 62: nit: not a particularly helpful comment.

src/main/java/bio/terra/janitor/service/pubsub/TrackedResourceSubscriber.java
```diff
@@ -0,0 +1,65 @@
...
+    if (!trackResourcePubsubConfiguration.isPubsubEnabled()) {
+      return;
+    }
+
+    createTrackedResourceSubscriber(
+        trackResourcePubsubConfiguration.getProjectId(),
+        trackResourcePubsubConfiguration.getSubscription());
+  }
+
+  private void createTrackedResourceSubscriber(String projectId, String subscriptionId) {
```
wchamber, line 40: seems a little funny to pull this into a separate method when it's the end of the function and there's nothing else to the initialize function.

src/main/java/bio/terra/janitor/common/exception/InvalidMessageException.java
```diff
@@ -0,0 +1,15 @@
+package bio.terra.janitor.common.exception;
+
+/**
+ * Exception when {@link bio.terra.generated.model.CreateResourceRequestBody} message is invalid.
+ */
+public class InvalidMessageException extends BadRequestException {
```
wchamber, line 6: why does this extend BadRequetException? The pub/sub queue isn't going to report to the ErrorModel of our API.

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+@TestPropertySource("classpath:application-integration-test.properties")
+public class TrackResourceIntegrationTest {
+  @Autowired private TrackResourcePubsubConfiguration trackResourcePubsubConfiguration;
+  @Autowired private TestConfiguration testConfiguration;
+  @Autowired private JanitorDao janitorDao;
+
+  private static Publisher publisher;
+
+  /** Google Service Account path used to publish message. */
+  private static String CLIENT_SERVICE_ACCOUNT_PATH = "rendered/client-sa-account.json";
```
wchamber, line 48: maybe add this to TestConfiguration? maybe include the function to create the GoogleCredentials for it as well there?

src/test/resources/application-integration-test.properties
```diff
@@ -0,0 +1,21 @@
+server.port=8080
```
wchamber, line 1: I think we should try to compose property files, only overriding some of the values between this integration-test file and the test/resources/application.properties file. I think Spring has a way to do that.

src/test/resources/application-integration-test.properties
```diff
@@ -0,0 +1,21 @@
...
+janitor.primary.schedulingEnabled=false
+stairway.maxParallelFlights=40
+stairway.migrateUpgrade=true
+stairway.forceCleanStart=true
+stairway.quietDownTimeout=20s
+stairway.terminateTimeout=5s
+pubsub.track-resource.pubsubEnabled=true
+pubsub.track-resource.projectId=terra-kernel-k8s
+pubsub.track-resource.subscription=crljanitor-toolsalpha-pubsub-sub
+test.trackResourceTopicId=crljanitor-toolsalpha-pubsub-topic
```
wchamber, line 21: where does this value come from? Does it need to be this value?

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+@ContextConfiguration(classes = Main.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@TestPropertySource("classpath:application-integration-test.properties")
+public class TrackResourceIntegrationTest {
+  @Autowired private TrackResourcePubsubConfiguration trackResourcePubsubConfiguration;
+  @Autowired private TestConfiguration testConfiguration;
+  @Autowired private JanitorDao janitorDao;
+
+  private static Publisher publisher;
```
wchamber, line 45: nit: I don't think this should be static

.github/workflows/build_and_test.yml
yonghaoy: Changed to toolsalpha

src/main/java/bio/terra/janitor/common/exception/InvalidMessageException.java
yonghaoy: ah, right. RuntimeE?

src/main/java/bio/terra/janitor/app/configuration/TrackResourcePubsubConfiguration.java
yonghaoy: Those are prefixes, and for the prefixes, it need to be with dashes. I got compliants if using mixed of lower$upper case as prefix.

.github/workflows/build_and_test.yml
yonghaoy: changed to test.yml.
I deleted the file and tried use git mv command, but it still can not get the git history. What would be the command to let git track the change history? Thanks

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
yonghaoy: TestConfiguration seems not the right place to put it. It is a `@ConfigurationProperties` classes in test, just like the configs in main/ files. But for the SA path, that seems not to be an application config(don't need to change very offten by different env) and will be changed along with the `render-config/sh file path change`? 
And for the `GoogleCredentials `, that seems not to be a config as well. We don't want to mix `@ConfigurationProperties` classes with `util` classes?

src/test/resources/application-integration-test.properties
yonghaoy: those values come from vault. It need to be some values which points to read Google cloud resource.

src/test/resources/application-integration-test.properties
yonghaoy: done

.github/workflows/build_and_test.yml
wchamber: probably not worth it then. If git mv doesn't do it, not sure what else I'd try.

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
wchamber: I don't see why the SA path is any less of an application config than the the trackResourceTopicId. These java classes don't know about render-config.sh putting things in the same place. It's fine to have a default value that matches the destination of render-config.sh, but if eels like if we have configuration properties, file paths are things that should be configurable.

If the SA path is in the `TestConfiguration` then turning the path into something useful for tests, a GoogleCredentials, seems reasonable to me. `getClientCredentialsOrDie`

src/test/resources/application-integration-test.properties
wchamber: Sorry, I don't really understand your reply. If these are supposed to come from vault, why are they hard coded here instead of rendered as config? As-is, these are magic values without comment.

wchamber:
per offline conversation, if we want to hard code things that are coming from vault, that's ok if we comment where we're copying from and a TODO some ticket to address eventually.

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -85,22 +84,26 @@ public void subscribeTrackResource() throws Exception {
...
-
-  private static ServiceAccountCredentials getGoogleCredentialsOrDie(String serviceAccountPath) {
-    try {
-      return ServiceAccountCredentials.fromStream(
-          Thread.currentThread().getContextClassLoader().getResourceAsStream(serviceAccountPath));
-    } catch (Exception e) {
-      throw new RuntimeException(
-          "Unable to load GoogleCredentials from " + serviceAccountPath + "\n", e);
-    }
+        Duration.between(trackedResourceInfo.getCreation(), trackedResourceInfo.getExpiration()));
```
wchamber, line 105: So now that we're using pub/sub, I don't think we should rely on the time we receive the message to account for the creation time & set the ttl. by the nature of pub/sub, that could be a long time after the resource was created. That was fine for the post endpoint, but I think we should change the format of the CreateResourceRequestBody to include the creation and expiration time now. Fine to do in a follow up PR, or now. this will be a breaking change to our api, so we'll have to fix up the CRL usage, but we have to change it anyway for pub/sub.

src/test/java/bio/terra/janitor/integration/common/configuration/TestConfiguration.java
```diff
@@ -11,11 +12,30 @@
...
+
+  public String getClientSaPath() {
+    return clientSaPath;
+  }
+
+  public void setClientSaPath(String clientSaPath) {
+    this.clientSaPath = clientSaPath;
+  }
+
+  public ServiceAccountCredentials getClientGoogleCredentialsOrDie() {
```
wchamber, line 33: add a comment on what "client" credentials are used for and what we expect them to have credentials to do.

src/test/java/bio/terra/janitor/integration/common/configuration/TestConfiguration.java
```diff
@@ -11,11 +12,30 @@
 public class TestConfiguration {
   private String trackResourceTopicId;
 
+  private String clientSaPath;
```
wchamber, line 15: nit: abbreviations are hard to understand later. Ok to spell out ServiceAccount.

src/test/resources/application-integration-test.properties
wchamber: Maybe add a link to the terraform module or deployment to where these are originating from. then LG.
Let's file a placeholder bug to TODO here.

build.gradle
```diff
@@ -141,3 +143,18 @@ task unitTest(type: Test) {
     outputs.upToDateWhen { false }
 }
 
+def googleCredentialsFile = "${projectDir}/src/test/resources/rendered/sa-account.json"
+
+task integrationTest(type: Test) {
+    // Sets the application Google Credentials.
+    environment = [
+            'GOOGLE_APPLICATION_CREDENTIALS': "${googleCredentialsFile}"
```
wchamber, line 151: We're going to need the application google credentials set for running the service as well. Can we start a service locally or in k8s with these changes as-is, or do we need to add environment variables to run_local.sh and the deployment.yaml?
Should running locally call render-config.sh?

render-config.sh
```diff
@@ -0,0 +1,16 @@
...
+VAULT_CLIENT_SERVICE_ACCOUNT_PATH=secret/dsde/terra/kernel/integration/toolsalpha/crl_janitor/client-sa
+SERVICE_ACCOUNT_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/rendered/sa-account.json
+CLIENT_SERVICE_ACCOUNT_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/rendered/client-sa-account.json
+
+docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
+            vault read -format json ${VAULT_SERVICE_ACCOUNT_PATH} \
+            | jq -r .data.key | base64 -d > ${SERVICE_ACCOUNT_OUTPUT_FILE_PATH}
+docker run --rm --cap-add IPC_LOCK \
+            -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
+            vault read -format json ${VAULT_SERVICE_ACCOUNT_PATH} \
```
wchamber, line 15: should this be VAULT_CLIENT_SERVICE_ACCOUNT?

render-config.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+VAULT_TOKEN=${1:-$(cat $HOME/.vault-token)}
+DSDE_TOOLBOX_DOCKER_IMAGE=broadinstitute/dsde-toolbox:consul-0.20.0
+VAULT_SERVICE_ACCOUNT_PATH=secret/dsde/terra/kernel/integration/toolsalpha/crl_janitor/app-sa
+VAULT_CLIENT_SERVICE_ACCOUNT_PATH=secret/dsde/terra/kernel/integration/toolsalpha/crl_janitor/client-sa
```
wchamber, line 6: Worth linking to the deployment repo where these are defined. I think we should also TODO for the same issue of wiring from terraform through to configs as the other TODO.

src/main/resources/application.properties
```diff
@@ -15,3 +15,6 @@ stairway.migrateUpgrade=true
 stairway.forceCleanStart=true
 stairway.quietDownTimeout=20s
 stairway.terminateTimeout=5s
+pubsub.track-resource.pubsubEnabled=${TRACK_RESOURCE_PUBSUB_ENABLED}
+pubsub.track-resource.projectId=${TRACK_RESOURCE_PUBSUB_TOPIC}
```
wchamber, line 19: the property is the project id, but the env variable is for the topic. It doesn't seem like the service uses the topic in the subscription at all. what's up with that?
the fact that this isn't caught be any test makes me think that we should change our test configuration to set these environment variables instead of overriding them in the application-integration-test.properties.

render-config.sh
yonghaoy: Done. The test passed because carjaniror-sa has project level pubsub admin permissions, so we can use that to publish message as well.
https://github.com/broadinstitute/terraform-ap-modules/blob/master/crl-janitor/variables.tf#L45
I saw that last week and planed to check with you on reason for this permission. I remembered you need it for some stairway pubsub stuff, but I don't know the detail

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
yonghaoy: I think we can get publish time from message. Can we use that? Changed to use that timestamp for now, and will have some follow up PRs if we want client to set the time. But I feel the publish time is fine to use?

build.gradle
yonghaoy: Local: we will need to modify the run_local.sh to set system environment variables.  I didn't put it because we turned off the pubsub in local server and we don't need this yet. 

K8s: I further think we should also add this into helm-chart it in k8s deployment to tell the service use the sa account. 
That is what I did for IC:
https://github.com/broadinstitute/terra-helm/blob/master/charts/identityconcentrator/templates/deployment.yaml#L35
Also, I checked the GKE log seems it is using the correct SA account somehow. But as we are adding more SA accounts and make them granular, we should have `GOOGLE_APPLICATION_CREDENTIALS ` in helm-chart.

render-config.sh
yonghaoy: Done

src/main/resources/application.properties
yonghaoy: ah, sorry, this should be the project id. It get messed up when I tried to add topic & project & subscription then deleted the topic. Didn't find it out because I haven't deployed this to k8s yet.  It works in test because in test we stetted it correctly there.  We don't need topic because subscription is created in Terraform.
Sorry, I didn't capture that because I havn't deploied that to my env yet, I can set those environment variable in gradle. Do you want me set all of them there, including db uri, password, username, stairway stuff or just pubsub stuff?

wchamber: Eh, I feel like we should have a test at some point that captures configuration errors like this. That could be the 'deploy to a ephemeral k8s environment integration test,' but earlier is often better. Maybe we should include all the environment variables set in our deployment.yaml as a part of gradle for the integration tests. That feels like it's a little closer to what's going on, but on the other hand, that's only for the integration tests, not the unit or local runs. WDYT?

local-dev/setup_gke_deploy.sh
```diff
@@ -9,8 +9,8 @@ set -e
 # Required input
 ENV=$1
 # Optional input
-TERRA_HELM_BRANCH=${2:-master}
-TERRA_HELMFILE_BRANCH=${3:-master}
+TERRA_HELM_BRANCH=${2:-yyu-PF-61}
+TERRA_HELMFILE_BRANCH=${3:-yyu-PF-64}
```
wchamber, line 13: don't submit this.

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
```diff
@@ -24,6 +24,12 @@ public JanitorService(JanitorDao janitorDao) {
 
   public CreatedResource createResource(CreateResourceRequestBody body) {
     Instant creationTime = Instant.now();
+    return createResourceWithCreation(body, creationTime);
+  }
+
+  /** Creates resources with given creation time. */
+  public CreatedResource createResourceWithCreation(
+      CreateResourceRequestBody body, Instant creationTime) {
```
wchamber, line 32: I think it's simpler to include the creation time within the CreateResourceRequetBody. Let's do that, in this PR or the next.

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
wchamber: Having an explicit creation time parameter seems neater to me. Also makes our tests a little easier to write.

yonghaoy: Done

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
yonghaoy: okay, I will do it in next PR

PR #33, opened 2020-08-03
PF-15 Add a get endpoint to retrieve tracked reosurces by CloudResourceUid

wchamber:
- Add yaml and implementation for getting tracked resources with a certain CloudResourceUid.
- Add ResultSetExtractor for getting the TrackedResource with all of its labels.

### DIFF
--- a/src/main/java/bio/terra/janitor/db/JanitorDao.java
+++ b/src/main/java/bio/terra/janitor/db/JanitorDao.java
@@ -7,9 +7,7 @@ import bio.terra.janitor.common.exception.InvalidResourceUidException;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableMap;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.time.Instant;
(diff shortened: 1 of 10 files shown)


src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -288,4 +339,33 @@ public static TrackedResourceAndFlight create(
       return new AutoValue_JanitorDao_TrackedResourceAndFlight(trackedResource, cleanupFlight);
     }
   }
+
+  /** A {@link TrackedResource} and the labels associated with it. */
+  @AutoValue
+  public abstract static class TrackedResourceAndLabels {
```
yonghaoy, line 345: Feels the db objects are used/defined everywhere now.
Shall we make it has a class like TrackedResource, Id, and State?

src/main/resources/api/service_openapi.yaml
```diff
@@ -72,6 +72,27 @@ paths:
...
           $ref: '#/components/responses/ErrorResponse'
+    get:
+      description: |
+        Return tracked resources matching the request filters
+      operationId: getResources
+      tags:
+        - janitor
+      parameters:
+        - name: cloudResourceUid
+          in: query
```
yonghaoy, line 83: Not sure what would be the best practice for defining a get for an object. In your solution, we treat the whole things as a string, not a query. And the rest api would be 

`get https://janitor/api/resource? cloudResourceUid=[a super long string]. `
Do we want to go with this solution or we want break down the cloudResourceUid and make it more like an 
"object". something like: 
`get https://janitor/api/resource? resourceType=[type]&projectName=[]&bucketName=[]`?
I don't know which is more preferred.

wchamber: I'm pretty strongly opposed to having another serialization format for a "cloud resource uid." I'm open to other singular formats. We have the use case of using it within a request body as a json object, and now in a get params (or path).

Doing this seemed not too bad. The string doesn't have to be too long if it's serialized to not include nulls, probably doesn't look more different than the individual params you're suggesting. It also gives us a single object to describe the "cloud resource uid" in the library and client apis rather than an ever expanding list of parameters.

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
```diff
@@ -60,6 +59,19 @@ public CreatedResource createResource(CreateResourceRequestBody body) {
     return Optional.of(createInfo(resource.get(), labels));
```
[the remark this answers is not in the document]
wchamber: done

src/main/java/bio/terra/janitor/db/JanitorDao.java
wchamber: done

PR #36, opened 2020-08-10
PF-3 Add stackdriver stats exporting and initial openecensus metrics for cleanup

wchamber:


wchamber:
Demo [metrics explorer link](https://console.cloud.google.com/monitoring/metrics-explorer?project=terra-kernel-k8s&pageState=%7B%22xyChart%22:%7B%22dataSets%22:%5B%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22custom.googleapis.com%2Fopencensus%2Fterra%2Fjanitor%2Fcleanup%2Fsubmission_duration%5C%22%20resource.type%3D%5C%22k8s_container%5C%22%22,%22minAlignmentPeriod%22:%2260s%22,%22aggregations%22:%5B%7B%22perSeriesAligner%22:%22ALIGN_DELTA%22,%22crossSeriesReducer%22:%22REDUCE_PERCENTILE_50%22,%22groupByFields%22:%5B%22resource.label.%5C%22cluster_name%5C%22%22,%22resource.label.%5C%22namespace_name%5C%22%22,%22resource.label.%5C%22pod_name%5C%22%22,%22resource.label.%5C%22container_name%5C%22%22%5D%7D,%7B%22crossSeriesReducer%22:%22REDUCE_NONE%22%7D%5D%7D,%22targetAxis%22:%22Y1%22,%22plotType%22:%22LINE%22%7D,%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22custom.googleapis.com%2Fopencensus%2Fterra%2Fjanitor%2Fcleanup%2Fcompletion_duration%5C%22%20resource.type%3D%5C%22k8s_container%5C%22%22,%22minAlignmentPeriod%22:%2260s%22,%22aggregations%22:%5B%7B%22perSeriesAligner%22:%22ALIGN_DELTA%22,%22crossSeriesReducer%22:%22REDUCE_PERCENTILE_50%22,%22groupByFields%22:%5B%22resource.label.%5C%22pod_name%5C%22%22,%22resource.label.%5C%22container_name%5C%22%22,%22resource.label.%5C%22cluster_name%5C%22%22,%22resource.label.%5C%22namespace_name%5C%22%22%5D%7D,%7B%22crossSeriesReducer%22:%22REDUCE_NONE%22%7D%5D%7D,%22targetAxis%22:%22Y1%22,%22plotType%22:%22LINE%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22constantLines%22:%5B%5D,%22timeshiftDuration%22:%220s%22,%22y1Axis%22:%7B%22label%22:%22y1Axis%22,%22scale%22:%22LINEAR%22%7D%7D,%22isAutoRefresh%22:true,%22timeSelection%22:%7B%22timeRange%22:%221h%22%7D%7D).

src/main/java/bio/terra/janitor/service/cleanup/MetricsHelper.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /** Record the duration of an attempt to submit a cleanup flight. */
+  public static void recordSubmissionDuration(Duration duration, boolean flightSubmitted) {
+    TagContext tctx =
+        TAGGER
+            .emptyBuilder()
+            .putLocal(KEY_SUCCESS, TagValue.create(Boolean.valueOf(flightSubmitted).toString()))
+            .build();
+    STATS_RECORDER.newMeasureMap().put(SUBMISSION_DURATION, duration.toMillis()).record(tctx);
```
yonghaoy, line 86: The CRL's code enters&exits the scope of the taxContext:
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/f47924513a11b65fc31b5529a67265b51c6a90ca/common/src/main/java/bio/terra/cloudres/util/MetricsHelper.java#L136

Do we also need to do this here? 
https://github.com/census-instrumentation/opencensus-java/blob/5be70440b53815eec1ab59513390aadbcec5cc9c/api/src/main/java/io/opencensus/tags/Tagger.java#L85

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -57,14 +60,18 @@ public FlightManager(
...
     Optional<TrackedResource> resource = updateResourceForCleaning(expiredBy, flightId);
     if (!resource.isPresent()) {
       // No resource to schedule.
       return Optional.empty();
     }
     // If submission fails, it will be recovered later.
-    submitToStairway(flightId, resource.get());
+    boolean submissionSuccessful = submitToStairway(flightId, resource.get());
+    // Only record duration of submission if there was something to attempt to schedule.
+    MetricsHelper.recordSubmissionDuration(
```
yonghaoy, line 73: Currently we only record the duration and the boolean. Is that enough? Do we want more information like the flight id, resource type, client id stuff for easier debug?

wchamber: These initial metrics are for how long the interaction between the JanitorDao and Stairway take, so I thought to only include what might be interesting there. I don't think the resource type or the client id will matter for those (and there's a limit on how many distinct values you should have in your metrics, so we really shouldn't include the flight id which is always unique. flight id is more for logs).
We'll still want to add more metrics about the resource type, client id, and tracked resource state, but not in this PR. I'm thinking we periodically query the database with some select-count-groupby and export that.

src/main/java/bio/terra/janitor/service/cleanup/MetricsHelper.java
wchamber: I don't think we need that here or in the CRL code. The scope allows the `tctx` to be implicit so that you can call record with no arguments, but we don't need it to be implicit and can just call `record(tctx)` explicitly. https://github.com/census-instrumentation/opencensus-java/blob/5be70440b53815eec1ab59513390aadbcec5cc9c/api/src/main/java/io/opencensus/stats/MeasureMap.java#L93-L102

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
mflinn-broad: Just wanted to add a +1 to what Will is saying here. When choosing labels to associate with a metric, it's really important to avoid using anything we know will exhibit high cardinality such as short-lived unique ids. This can cause performance issues when querying metrics. This can also cause the number of unique time series tracked in stackdriver to massively increase which could lead to quota/billing issues. So I think this is a good thing to keep in mind when selecting metrics labels.

src/main/java/bio/terra/janitor/service/stackdriver/StatsExporter.java
```diff
@@ -0,0 +1,34 @@
...
+    this.stackdriverConfiguration = stackdriverConfiguration;
+  }
+
+  public void initialize() {
+    logger.info("Stackdriver stats enabled: {}.", stackdriverConfiguration.isEnabled());
+    if (!stackdriverConfiguration.isEnabled()) {
+      return;
+    }
+    try {
+      StackdriverStatsExporter.createAndRegister();
```
mflinn-broad, line 29: If opencensus doesn't have support for automatic parsing of a version. One way to add a version tag to all metrics would be to pass an instance of `StackdriverStatsConfiguration` to `createAndRegister`.
The `setConstantLabels` method [here](https://www.javadoc.io/doc/io.opencensus/opencensus-exporter-stats-stackdriver/latest/io/opencensus/exporter/stats/stackdriver/StackdriverStatsConfiguration.Builder.html) could be used to automatically attach a version label to all metrics

wchamber: We talked about this offline. Right now, the image value has shown up at least once in stackdriver. Long term, we do want to be able to distinguish metrics by the image, but it'd be nice if we can include it as metadata on the resource instead of included in the metric labels. We should find a standard way to do this long term, and set images on each label if we have to.

PR #40, opened 2020-08-13
[PF-29] Support google bucket cleanup

yonghaoy:


build.gradle
```diff
@@ -6,6 +6,7 @@ buildscript {
 
 plugins {
     id 'java'
+    id 'java-library'
```
wchamber, line 9: it seems funny to me that we need to build a java-library since this is a java service binary. what do we need this for?

.github/workflows/test.yml
```diff
@@ -47,21 +47,10 @@ jobs:
...
-            vault read -format json $VAULT_SERVICE_ACCOUNT_PATH \
-            | jq -r .data.key | base64 -d > $SERVICE_ACCOUNT_FILE &&#
-          docker run --rm --cap-add IPC_LOCK \
-            -e "VAULT_TOKEN=${{ steps.vault-token-step.outputs.vault-token }}" \
-            -e "VAULT_ADDR=${VAULT_ADDR}" \
-            vault:1.1.0 \
-            vault read -format json $VAULT_CLIENT_SERVICE_ACCOUNT_PATH \
-            | jq -r .data.key | base64 -d > $SERVICE_ACCOUNT_CLIENT_FILE
+      - name: Grant execute permission for render-config
+        run: chmod +x render-config.sh
```
wchamber, line 51: I think render-config.sh is now also a critical part of local-dev. Can we (probably in a separate PR) add some documentation to the README about using it? should it live in the local-dev directory?

src/main/java/bio/terra/janitor/common/CredentialUtils.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.janitor.common;
+
+import bio.terra.janitor.common.exception.JanitorCredentialException;
+import com.google.auth.oauth2.ServiceAccountCredentials;
+
+/** Utilities for handling credentials. */
+public class CredentialUtils {
+  /** Gets the {@link ServiceAccountCredentials} from a file path. */
+  public static ServiceAccountCredentials getGoogleCredentialsOrDie(String serviceAccountPath) {
```
wchamber, line 9: Do we need this in main since we're setting up environment variables to make the service account use the default google credentials? I see the utility of this in test where we have client service accounts, so let's move it to src/test?

src/main/java/bio/terra/janitor/common/CredentialUtils.java
```diff
@@ -0,0 +1,18 @@
...
+    try {
+      return ServiceAccountCredentials.fromStream(
+          Thread.currentThread().getContextClassLoader().getResourceAsStream(serviceAccountPath));
+    } catch (Exception e) {
+      throw new JanitorCredentialException(
```
wchamber, line 14: I think having a special exception class for this is overkill. This only happens initially at startup, so I'm not worried about trying to use a new class to trace it down.

src/main/java/bio/terra/janitor/db/TrackedResource.java
```diff
@@ -9,15 +14,24 @@
  * table in the Janitor's database.
  */
 @AutoValue
+@JsonSerialize(as = TrackedResource.class)
+@JsonDeserialize(builder = AutoValue_TrackedResource.Builder.class)
```
wchamber, line 18: I don't think we should serialize the whole TrackedResource class into the FlightParamters. The cleanup flights don't need to know things like creation/expiration, or the TrackedResourceState that it starts in.
I think we should make a dedicated POJO class that we can serialize as JSON for CleanupFlights. That also allows us to add things to diverge our standard cleanup flight parameters from what's in TrackedResourceState. Keeping serialization (and Stairway counts as a serialization) as a separate layer from our internal representation class (TrackedResource) is a good place to start.

src/main/java/bio/terra/janitor/service/cleanup/flight/FlightMapKeys.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.janitor.service.cleanup.flight;
+
+/** Keys for Janitor's stairway flight input parameter map. */
+public final class FlightMapKeys {
+  private FlightMapKeys() {}
+
+  public static final String TRACKED_RESOURCE = "trackedResource";
```
wchamber, line 7: Once we have a dedicated serialization class for a cleanup flight parameters, I think that class should include helper methods to set/get from FlightMap.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.cloud.storage.StorageOptions;
+
+public class GoogleBucketCleanupStep implements Step {
+  private final StorageCow storageCow;
+
+  public GoogleBucketCleanupStep() {
+    // Janitor only uses CRL Cows to delete resources. Cleanup is not needed.
+    this.storageCow =
+        new StorageCow(
+            ClientConfig.Builder.newBuilder().setClient("terra-crl-janitor").build(),
```
wchamber, line 18: Maybe doesn't have to be this PR, but I think long term this (and other cleanup steps) should take the ClientConfig as an argument to the constructor. Then, we should make a `@Component` that creates a ClientConfig and grab that during Flight construction.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,36 @@
...
+            StorageOptions.getDefaultInstance());
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    TrackedResource trackedResource =
+        flightContext
+            .getInputParameters()
+            .get(FlightMapKeys.TRACKED_RESOURCE, TrackedResource.class);
+    storageCow.delete(trackedResource.cloudResourceUid().getGoogleBucketUid().getBucketName());
```
wchamber, line 28: To delete a bucket, we need to delete all the objects in the bucket. We can also leave a TODO for now. Let's start with a naive iteration over the blobs, deleting them individually with a `TODO: Set bucket lifetime to 0 and wait once Stairway supports waiting. This is more robust for buckets with many or very large objects.`

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,36 @@
...
+  public GoogleBucketCleanupStep() {
+    // Janitor only uses CRL Cows to delete resources. Cleanup is not needed.
+    this.storageCow =
+        new StorageCow(
+            ClientConfig.Builder.newBuilder().setClient("terra-crl-janitor").build(),
+            StorageOptions.getDefaultInstance());
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
```
wchamber, line 23: our do cleanup steps need to check a number of conditions and respond accordingly.
We can do this in a follow up PR if you like, but leave a TODO at least.
https://docs.google.com/document/d/13mYVJML_fOLsX1gxQxRJgECUNT27dAMKzEJxUEvtQqM/edit#heading=h.p8sy8wh5t9mj

src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
```diff
@@ -32,4 +40,16 @@ public UnsupportedCleanupFlight(FlightMap inputParameters, Object applicationCon
...
   }
+
+  /** A Flight to cleanup a GoogleBucket. */
+  public static class GoogleBucketCleanupFlight extends Flight {
+    public GoogleBucketCleanupFlight(FlightMap inputParameters, Object applicationContext) {
+      super(inputParameters, applicationContext);
+      JanitorDao janitorDao =
+          ((ApplicationContext) applicationContext).getBean("janitorDao", JanitorDao.class);
+      addStep(new InitialCleanupStep(janitorDao));
+      addStep(new GoogleBucketCleanupStep());
```
wchamber, line 51: we should add some retries to the cleanup step.

src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
```diff
@@ -32,4 +40,16 @@ public UnsupportedCleanupFlight(FlightMap inputParameters, Object applicationCon
       addStep(new FinalCleanupStep(janitorDao));
     }
   }
+
+  /** A Flight to cleanup a GoogleBucket. */
+  public static class GoogleBucketCleanupFlight extends Flight {
```
wchamber, line 45: nit: maybe split out the cleanup flights into their own files? I think it might be nice to have a subdirectory `service/cleanup/flights` to hold all of our Flights and Steps.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,36 @@
...
+
+import bio.terra.cloudres.common.ClientConfig;
+import bio.terra.cloudres.google.storage.StorageCow;
+import bio.terra.janitor.db.TrackedResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import com.google.cloud.storage.StorageOptions;
+
+public class GoogleBucketCleanupStep implements Step {
```
wchamber, line 11: add a javadoc comment

src/test/java/bio/terra/janitor/common/configuration/TestConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+@Component
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "test")
+public class TestConfiguration {
+  /** How long to keep the resource before the 'prod' Janitor do the cleanup. */
+  public static Duration RESOURCE_TIME_TO_LIVE = Duration.ofMinutes(30);
+
+  private String resourceProjectId;
+  private String prodTrackResourceTopicId;
+  private String prodCrlClientCredentialFilePath;
```
wchamber, line 23: we probably don't want to use 'prod.' Is this the Crl client or the Janitor client?

these need javadoc comments

src/test/java/bio/terra/janitor/common/configuration/TestConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+
+  private String resourceCredentialFilePath;
```
wchamber, line 25: say more.

src/test/java/bio/terra/janitor/common/configuration/TestConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+  public void setProdTrackResourceTopicId(String prodTrackResourceTopicId) {
+    this.prodTrackResourceTopicId = prodTrackResourceTopicId;
+  }
+
+  public void setResourceCredentialFilePath(String resourceCredentialFilePath) {
+    this.resourceCredentialFilePath = resourceCredentialFilePath;
+  }
+
+  /** Creates {@link ClientConfig} for using CRL in test. */
+  public ClientConfig getCrlClientConfig() {
```
wchamber, line 57: nit: 'create' instead of 'get' since it makes a new one. And 'Crl' is redundant since that's implied by the package name of the type.

src/test/java/bio/terra/janitor/integration/common/configuration/IntegrationTestConfiguration.java
```diff
@@ -8,8 +10,8 @@
 /** Configs in integration test. */
 @Component
 @EnableConfigurationProperties
-@ConfigurationProperties(prefix = "test")
-public class TestConfiguration {
+@ConfigurationProperties(prefix = "it")
+public class IntegrationTestConfiguration {
```
wchamber, line 14: Why do we have this class and the TestConfiguration class?

src/test/java/bio/terra/janitor/service/cleanup/FlightSchedulerTest.java
```diff
@@ -75,7 +95,19 @@ private boolean resourceStateIs(TrackedResourceId id, TrackedResourceState expec
   }
 
   @Test
-  public void resourceScheduledForCleanup() throws Exception {
+  public void resourceScheduledForCleanup_googleBucket() throws Exception {
```
wchamber, line 98: I don't think we should be changing the FlightSchedulerTest for buckets.
The scheduler test is about handing things off to Stairway and back. It's unit testing that integration. We don't need to involve real flights for real resources.
We can have a test for the individual cleanup steps, or the whole flight, but we don't need to involve the scheduler to do them.
We can have an end to end test using the Janitor's public APIs to test the whole thing.

src/test/java/bio/terra/janitor/integration/common/configuration/IntegrationTestConfiguration.java
```diff
@@ -8,8 +10,8 @@
 /** Configs in integration test. */
 @Component
 @EnableConfigurationProperties
-@ConfigurationProperties(prefix = "test")
-public class TestConfiguration {
+@ConfigurationProperties(prefix = "it")
```
wchamber, line 13: "it" is not a very descriptive prefix. Prefer avoiding abbreviations.

yonghaoy: I thought it means for 'integrationTest' and this is how WM names it:
https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/test/resources/application-integration-test.properties#L1

Any other recommandations?full name like: integrationTest?

wchamber: yeah, "integrationTest" full name.

src/test/java/bio/terra/janitor/service/cleanup/FlightSchedulerTest.java
yonghaoy: Those tests are not there, those lives under FlightManager before. They were at FlightManagerTests where I tested that flight. 
Then I found this test failed because it expects state to be ERROR but now they are DONE. So I moved the test here. 
Do you want me to move the cleanup test back to FlightMangaerTest or integration test?

src/test/java/bio/terra/janitor/integration/common/configuration/IntegrationTestConfiguration.java
yonghaoy: TestConfiguration is for unit test configs. And integrationTestIntegration is for integration test.

src/main/java/bio/terra/janitor/db/TrackedResource.java
yonghaoy: I think just CloudResourceUid is enough for a flight to cleanup the resource. So just use the cloudResourceUid instead?

src/main/java/bio/terra/janitor/common/CredentialUtils.java
yonghaoy: Done

src/main/java/bio/terra/janitor/common/CredentialUtils.java
yonghaoy: Done

.github/workflows/test.yml
yonghaoy: I don't feel it is right to move this into local-dev. Because we also need this to run test. 
It is wired that we run local-dev/render-config.sh before running test locally and in github. We should just let the local-dev to run ../render-config.sh first then set the environment variable. WDYT? I will have a seperate PR for that

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
yonghaoy: Done

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
yonghaoy: Added a TODO

src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
yonghaoy: The benefit of using CRL is it already handles retries for us. Do we want to add more retry logic on top of that?

src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
yonghaoy: All steps are already under `service/cleanup/flights`. Also moved to flight there

src/main/java/bio/terra/janitor/db/TrackedResource.java
yonghaoy: Added CleanupParams

src/main/java/bio/terra/janitor/service/cleanup/flight/FlightMapKeys.java
yonghaoy: done

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
yonghaoy: done

src/test/java/bio/terra/janitor/common/configuration/TestConfiguration.java
yonghaoy: I feel prod the the word most close to the use case. 
We will need to publish two copies of messages: One for test(toolsalpha), one to the tools env, I call the tools env to prod because I feel that is the 'prod' for janitor. Is there any other name I can use to differentiate them? Thanks

.github/workflows/test.yml
wchamber: eh, when do we run tests except for "locally." even when the github action runs the test, it runs it "locally" in its container, just from a known clean state. But up to you.

README.md
```diff
@@ -12,6 +12,13 @@ done by multiple instances should be confined to the primary.
 
 # Development
 
+## Configs Rendering
+Local Testing and Github Action tests require credentials to be able to call GCP, run
+``` render-config.sh``` first to local testing. It generates:
```
wchamber, line 17: nit: 'first for local testing'

build.gradle
```diff
@@ -132,11 +135,21 @@ sourceSets.main.java.srcDir "${openapiTargetDir}/src/main/java"
...
+    environment = [
+            'GOOGLE_APPLICATION_CREDENTIALS': "${googleCredentialsFile}"
+    ]
     useJUnitPlatform()
 }
 
 task unitTest(type: Test) {
+    // Sets the application Google Credentials.
+    environment = [
+            'GOOGLE_APPLICATION_CREDENTIALS': "${googleCredentialsFile}"
```
wchamber, line 151: why do our unitTests need google credentials? Shouldn't only integrationTets need the service account credentials?

build.gradle
```diff
@@ -132,11 +135,21 @@ sourceSets.main.java.srcDir "${openapiTargetDir}/src/main/java"
 ideaModule.dependsOn swaggerSources.server.code
 // end of OpenAPI Server Generation
 
+def googleCredentialsFile = "${projectDir}/src/test/resources/rendered/sa-account.json"
```
wchamber, line 138: nit: add a comment `// This is the path to the default Google service account for the janitor service to run as. `

src/main/java/bio/terra/janitor/db/TrackedResourceId.java
```diff
@@ -1,15 +1,24 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.auto.value.AutoValue;
 import java.util.UUID;
 
 /** Wraps the tracked_resource_id in db tracked_resource table. */
 @AutoValue
+@JsonSerialize(as = TrackedResourceId.class)
```
wchamber, line 13: add tests for json serialization of TrackedResourceId.

src/main/java/bio/terra/janitor/service/cleanup/CleanupParams.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.janitor.service.cleanup;
+
+import bio.terra.generated.model.CloudResourceUid;
+import bio.terra.janitor.db.TrackedResourceId;
+import bio.terra.stairway.FlightMap;
+
+/** /** Helper class to manage input parameters in a {@link FlightMap}. */
```
wchamber, line 7: remove duplicate `/**` and add `in a {@link FlightMap} for all cleanup flights. */`

src/main/java/bio/terra/janitor/service/cleanup/CleanupParams.java
```diff
@@ -0,0 +1,25 @@
...
+public final class CleanupParams {
+  public static final String TRACKED_RESOURCE_ID = "trackedResourceId";
+  public static final String CLOUD_RESOURCE_UID = "cloudResourceUid";
```
wchamber, line 10: I think these should be private and we should have `set` methods.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupFlight.java
```diff
@@ -0,0 +1,25 @@
...
+public class GoogleBucketCleanupFlight extends Flight {
+  public GoogleBucketCleanupFlight(FlightMap inputParameters, Object applicationContext) {
+    super(inputParameters, applicationContext);
+    JanitorDao janitorDao =
+        ((ApplicationContext) applicationContext).getBean("janitorDao", JanitorDao.class);
+    ClientConfig clientConfig =
+        ((ApplicationContext) applicationContext).getBean("crlClientConfig", ClientConfig.class);
+    addStep(new InitialCleanupStep(janitorDao));
+    // Retry every 180 seconds for 5 times.
+    RetryRuleFixedInterval retryRule = new RetryRuleFixedInterval(180, 5);
```
wchamber, line 20: nit: use parameters comment instead. `new RetryRuleFixedInterval(/* intervalSeconds =*/ 180, /* maxCount =*/5);`

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,74 @@
...
+/** Step to cleanup Google Bucket resource. */
+public class GoogleBucketCleanupStep implements Step {
+  private Logger logger = LoggerFactory.getLogger(GoogleBucketCleanupStep.class);
+
+  private final StorageCow storageCow;
+  private final JanitorDao janitorDao;
+
+  public GoogleBucketCleanupStep(ClientConfig clientConfig, JanitorDao janitorDao) {
+    this.janitorDao = janitorDao;
+    // Janitor only uses CRL Cows to delete resources. Cleanup is not needed.
```
wchamber, line 30: comment's not appropriate here. this cleanupstep does not care how the ClientConfig is created.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,74 @@
...
+    Optional<TrackedResource> trackedResource =
+        janitorDao.retrieveTrackedResource(trackedResourceId);
+    // TrackedResource doesn't exist, this should not happen, fail the flight.
+    if (!trackedResource.isPresent()) {
+      logger.warn("trackedResource {} doesn't exist", trackedResourceId.toString());
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
+    }
+    TrackedResourceState state = trackedResource.get().trackedResourceState();
+    // State is DONE or ERROR, this should not happen, fail the flight.
+    if (state.equals(TrackedResourceState.DONE) || state.equals(TrackedResourceState.ERROR)) {
```
wchamber, line 51: let's whitelist the expected states instead of blacklisting unexpected states. Better yet, use an exhaustive switch statement so that this will not compile if we do not deal with all possible TrackedResourceStates.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,74 @@
...
+      logger.warn("Unexpected trackedResource state for {}.", trackedResourceId.toString());
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
+    }
+    // State is DUPLICATED or ABANDONED, this should not happen, complete the flight.
+    if (state.equals(TrackedResourceState.DUPLICATED)
+        || state.equals(TrackedResourceState.ABANDONED)) {
+      logger.warn("Unexpected trackedResource state for {}.", trackedResourceId.toString());
+      return StepResult.getStepResultSuccess();
+    }
```
wchamber, line 60: soon this should be shared code. Either move it somewhere to be re-used now, or leave a TODO saying we should pull it out and re-use it in other resource cleanup steps.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,74 @@
...
+    TrackedResourceState state = trackedResource.get().trackedResourceState();
+    // State is DONE or ERROR, this should not happen, fail the flight.
+    if (state.equals(TrackedResourceState.DONE) || state.equals(TrackedResourceState.ERROR)) {
+      logger.warn("Unexpected trackedResource state for {}.", trackedResourceId.toString());
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
+    }
+    // State is DUPLICATED or ABANDONED, this should not happen, complete the flight.
+    if (state.equals(TrackedResourceState.DUPLICATED)
+        || state.equals(TrackedResourceState.ABANDONED)) {
+      logger.warn("Unexpected trackedResource state for {}.", trackedResourceId.toString());
```
wchamber, line 58: DUPLICATE and ABANDONED are allowed states, not unexpected. fix comments and warning log. They do indicate a race with cleanup, so it's worth logging at warn.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,74 @@
...
+      return StepResult.getStepResultSuccess();
+    }
+    try {
+      storageCow.delete(cloudResourceUid.getGoogleBucketUid().getBucketName());
+      return StepResult.getStepResultSuccess();
+    } catch (StorageException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+      // Catch all exceptions from GOOGLE and consider this retryable error.
```
wchamber, line 66: nit: move the comment above the `return new StepResult`

src/test/java/bio/terra/janitor/common/CredentialUtils.java
```diff
@@ -0,0 +1,17 @@
...
+/** Utilities for handling credentials. */
+public class CredentialUtils {
+  /** Gets the {@link ServiceAccountCredentials} from a file path. */
+  public static ServiceAccountCredentials getGoogleCredentialsOrDie(String serviceAccountPath) {
+    try {
+      return ServiceAccountCredentials.fromStream(
+          Thread.currentThread().getContextClassLoader().getResourceAsStream(serviceAccountPath));
+    } catch (Exception e) {
+      throw new RuntimeException(
+          "Unable to load GoogleCredentials from configuration" + serviceAccountPath + "\n", e);
```
wchamber, line 14: nit: don't include the "\n" in the message.

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -84,18 +84,32 @@ public void tearDownPubsub() throws Exception {
...
 
   @Test
-  public void subscribeTrackResource() throws Exception {
+  public void subscribeAndCleanupResource_googleBucket() throws Exception {
+    // Creates bucket and verify.
+    String bucketName = UUID.randomUUID().toString();
+    assertNull(storageCow.get(bucketName));
+    BucketCow createdBucket = storageCow.create(BucketInfo.of(bucketName));
+    assertEquals(bucketName, createdBucket.getBucketInfo().getName());
+    assertEquals(bucketName, storageCow.get(bucketName).getBucketInfo().getName());
```
wchamber, line 93: Let's just do one of the asserts that the bucket exists, not both.

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -106,10 +120,22 @@ public void subscribeTrackResource() throws Exception {
...
+    assertEquals(resource, trackedResourceInfo.getResourceUid());
+    assertEquals(publishTime, trackedResourceInfo.getCreation());
+    assertEquals(publishTime, trackedResourceInfo.getExpiration());
     assertEquals(DEFAULT_LABELS, trackedResourceInfo.getLabels());
+
+    // Resource is removed
+    assertNull(storageCow.get(bucketName));
+  }
+
+  /** Returns a new {@link CreateResourceRequestBody} that is ready for cleanup. */
```
wchamber, line 132: nit: `} for a resource that is ready for cleanup. */`

src/test/java/bio/terra/janitor/integration/common/configuration/TestConfiguration.java
```diff
@@ -10,9 +15,28 @@
...
+
+  /** pubsub topic id to publish track resource to Janitor test env(toolsalpha) */
   private String trackResourceTopicId;
 
-  private String clientServiceAccountPath;
+  /** pubsub topic id to publish track resource to Janitor prod env(tools) */
+  private String prodTrackResourceTopicId;
+
+  /** Credential file path to be able to pubish message to Janitor test env(toolsalpha). */
+  private String janitorClientServiceAccountPath;
```
wchamber, line 34: wdyt about having a nested class for the pubsub project id, topic id, and service account to use for publishing? I think those three fields ar etied together and it'd be clearer to have them grouped  with an instance for test env and prod env.
```
public class TrackPubSubConfiguration {
  private String projectId;
  private String topicId;
  private String clientServiceAccountPath;

  public GoogleCredentials getClientCredentialsOrDie() { ... }
}
```

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -0,0 +1,74 @@
...
+    this.janitorDao = janitorDao;
+    // Janitor only uses CRL Cows to delete resources. Cleanup is not needed.
+    this.storageCow = new StorageCow(clientConfig, StorageOptions.getDefaultInstance());
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    // TODO(yonghao): Set bucket lifetime to 0 and wait once Stairway supports waiting. This is more
+    // robust for buckets
+    //  with many or very large objects.
```
wchamber, line 38: even before we have Stairway waiting, we still need to delete objects within a bucket to delete a bucket.
Fine to `TODO(PF-29): Iterate through and delete all objects within the bucket.`, but PF-29 isn't done until we can delete a bucket that has an object in it.
Also move these TODO comments to the storageCow.delete, instead of all the way up here.

.github/workflows/test.yml
yonghaoy: okay, moved to local-dev/

src/main/java/bio/terra/janitor/db/TrackedResourceId.java
yonghaoy: done

src/main/java/bio/terra/janitor/service/cleanup/CleanupParams.java
wchamber: I'd say let's use it for `get` (in the steps) and `set` (in flight submission construction) or not at all. Either we're encapsulating these flightmap string keys, or we're not.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupFlight.java
```diff
@@ -0,0 +1,26 @@
...
+    super(inputParameters, applicationContext);
+    JanitorDao janitorDao =
+        ((ApplicationContext) applicationContext).getBean("janitorDao", JanitorDao.class);
+    ClientConfig clientConfig =
+        ((ApplicationContext) applicationContext).getBean("crlClientConfig", ClientConfig.class);
+    addStep(new InitialCleanupStep(janitorDao));
+
+    RetryRuleFixedInterval retryRule =
+        new RetryRuleFixedInterval(/* intervalSeconds =*/ 180, /* maxCount =*/ 5);
+
```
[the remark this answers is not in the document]
yonghaoy: done

src/main/java/bio/terra/janitor/service/cleanup/flight/ResourceCleanupStep.java
```diff
@@ -0,0 +1,80 @@
...
+    // TODO(PF-29): Iterate through and delete all objects within the bucket.
+    CleanupParams cleanupParams = new CleanupParams(flightContext.getInputParameters());
+    CloudResourceUid cloudResourceUid = cleanupParams.getResourceUid();
+    TrackedResourceId trackedResourceId = cleanupParams.getTrackedResourceId();
+    Optional<TrackedResource> trackedResource =
+        janitorDao.retrieveTrackedResource(trackedResourceId);
+    // TrackedResource doesn't exist, this should not happen, fail the flight.
+    if (!trackedResource.isPresent()) {
+      logger.warn("trackedResource {} doesn't exist", trackedResourceId.toString());
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
```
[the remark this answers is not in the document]
yonghaoy: Then I will just use if(cleaning){cleanUp} else {illegalstateexception}

src/test/java/bio/terra/janitor/db/TrackedResourceIdTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.UUID;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("unit")
+public class TrackedResourceIdTest {
+  @Test
+  public void serialize() throws Exception {
```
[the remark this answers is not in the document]
yonghaoy: Done

src/main/java/bio/terra/janitor/service/cleanup/flight/ResourceCleanupStep.java
```diff
@@ -42,31 +41,13 @@ public StepResult doStep(FlightContext flightContext) {
...
-            String.format(
-                "Unsupported trackedResource state for trackedResource %s, state is %s",
-                trackedResourceId.toString(), state.toString()));
+    if (state.equals(CLEANING)) {
+      return cleanUp(cloudResourceUid);
+    } else {
+      throw new IllegalStateException(
+          String.format(
+              "Illegal trackedResource state for trackedResource %s, state is %s",
+              trackedResourceId.toString(), state.toString()));
```
wchamber, line 50: DUPLICATED and ABANDONED are not unsupported states. They are expected. They should result in stepresultsuccess.

src/main/java/bio/terra/janitor/service/cleanup/flight/ResourceCleanupStep.java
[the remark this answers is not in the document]
yonghaoy: Ah, sorry, I thought your comments are on the DUPLICATE, and ABANDON line

src/main/java/bio/terra/janitor/service/cleanup/flight/ResourceCleanupStep.java
yonghaoy: Done

PR #43, opened 2020-08-18
PF-5 Add logic for duplicating resources when they are created

wchamber:
- Use TrackedResourceFilter for retrieving TrackedResources with different filters in JanitorDao.

### DIFF
--- a/src/main/java/bio/terra/janitor/db/TrackedResourceFilter.java
+++ b/src/main/java/bio/terra/janitor/db/TrackedResourceFilter.java
@@ -5,6 +5,7 @@ import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableSet;
 import java.time.Instant;
 import java.util.Optional;
+import java.util.OptionalInt;
 import java.util.Set;
 
 /** A value class for filtering which resources are retrieved. */
@@ -22,11 +23,14 @@ public abstract class TrackedResourceFilter {
   /** If present, only resources with an expiration date less than or equal to this are allowed. */
   public abstract Optional<Instant> expiredBy();
 
+  /** If present, only return up to {@code limit} resources. */
+  public abstract OptionalInt limit();
+
   /** Creates a new builder that allows all resources. */
   public static Builder builder() {
     return new AutoValue_TrackedResourceFilter.Builder()
-            .allowedStates(ImmutableSet.of())
-            .forbiddenStates(ImmutableSet.of());
+        .allowedStates(ImmutableSet.of())
+        .forbiddenStates(ImmutableSet.of());
   }
 
   @AutoValue.Builder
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
```diff
@@ -2,37 +2,92 @@
...
+          duplicateResources.stream()
+              .map(TrackedResource::expiration)
+              .max(Instant::compareTo)
+              .get();
+      if (resource.expiration().isAfter(lastExpiration)) {
+        // The new resource expires after existing resources. The other resources are now
+        // duplicates of the new resource.
+        for (TrackedResource duplicateResource : duplicateResources) {
+          janitorDao.updateResourceState(
+              duplicateResource.trackedResourceId(), TrackedResourceState.DUPLICATED);
```
yonghaoy, line 80: Do you think ABANDONED should go with the CloudResourceUid or  trackedResourceId?  Shall we mark the new resource as abandon and previous one as DUPLICATED? Or just assume ABANDON is per trackedResourceId thing?

src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -91,17 +92,27 @@ public FlightManager(
...
+        janitorDao.retrieveResourcesMatching(
+            TrackedResourceFilter.builder()
+                .allowedStates(ImmutableSet.of(TrackedResourceState.READY))
+                .expiredBy(expiredBy)
+                .limit(1)
+                .build());
+    if (resources.isEmpty()) {
       return Optional.empty();
     }
+    if (resources.size() > 1) {
```
yonghaoy, line 105: I think you have `limit(1)` in the filter, and it will never reach here. Maybe remove the limit?

wchamber: I want the limit so that this only processes one resource at a time. At the same time, I have this check to make sure that it's behaving as we expect, otherwise there's a bug and something is broken. Easier to fail early.

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
wchamber: I think it should go with the trackedResourceId. Just because something was abandoned once doesn't mean we want any subsequent resources to also be abandoned.

src/main/java/bio/terra/janitor/db/TrackedResourceFilter.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public abstract Builder forbiddenStates(Set<TrackedResourceState> forbiddenStates);
+
+    public abstract Builder cloudResourceUid(CloudResourceUid cloudResourceUid);
+
+    public abstract Builder expiredBy(Instant expiredBy);
+
+    public abstract Builder limit(int value);
+
+    public abstract TrackedResourceFilter build();
```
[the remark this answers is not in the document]
wchamber: done.

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
yonghaoy: Got you. Might want to check with you the abandon endpoint looks like before you left. If we may choose to pull that ticket next week

PR #44, opened 2020-08-19
[PF-29] Cleanup Blobs during Bucket Cleanup

yonghaoy:


src/main/java/bio/terra/janitor/service/cleanup/flight/FinalCleanupStep.java
```diff
@@ -19,6 +19,8 @@ public FinalCleanupStep(JanitorDao janitorDao) {
 
   @Override
   public StepResult doStep(FlightContext flightContext) {
+    System.out.println("~~~~~~~ ENTERING FINISHING Stage");
+    System.out.println("~~~~~~~");
```
wchamber, line 23: remove debug prints before submitting

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -21,11 +24,19 @@ public GoogleBucketCleanupStep(ClientConfig clientConfig, JanitorDao janitorDao)
...
-      storageCow.delete(resourceUid.getGoogleBucketUid().getBucketName());
+      System.out.println("~~~~~~COMPLETE START");
+      Stopwatch stopwatch = Stopwatch.createStarted();
+      // Delete all Blobs before deleting bucket.
+      String bucketName = resourceUid.getGoogleBucketUid().getBucketName();
+      BucketCow bucketCow = storageCow.get(bucketName);
+      bucketCow.list().iterateAll().forEach(BlobCow::delete);
+      bucketCow.delete();
+      System.out.println("~~~~~~COMPLETE CLEANUP");
+      System.out.println(stopwatch.elapsed().abs());
```
wchamber, line 36: ditto remove debug prints and unused stopwatch?

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBucketCleanupStep.java
```diff
@@ -21,11 +24,19 @@ public GoogleBucketCleanupStep(ClientConfig clientConfig, JanitorDao janitorDao)
...
     // robust for buckets with many or very large objects.
-    // TODO(PF-29): Iterate through and delete all objects within the bucket.
     try {
-      storageCow.delete(resourceUid.getGoogleBucketUid().getBucketName());
+      System.out.println("~~~~~~COMPLETE START");
+      Stopwatch stopwatch = Stopwatch.createStarted();
+      // Delete all Blobs before deleting bucket.
+      String bucketName = resourceUid.getGoogleBucketUid().getBucketName();
+      BucketCow bucketCow = storageCow.get(bucketName);
+      bucketCow.list().iterateAll().forEach(BlobCow::delete);
```
wchamber, line 33: iterating over something while deleting them can have issues. Does this work? Should we instead collect all of the blob ids, then delete all of them? That won't work well if there are so many blobs that it OOMs us, but short of that, it seems a little safer. If we have that many blobs in a bucket, we may want to use the lifetime deletion then.
WDYT?

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -86,10 +94,14 @@ public void tearDownPubsub() throws Exception {
   @Test
   public void subscribeAndCleanupResource_googleBucket() throws Exception {
     // Creates bucket and verify.
-    String bucketName = UUID.randomUUID().toString();
+    String bucketName = randomName();
     assertNull(storageCow.get(bucketName));
-    storageCow.create(BucketInfo.of(bucketName));
+    BucketCow bucketCow = storageCow.create(BucketInfo.of(bucketName));
+    BlobId blobId = BlobId.of(bucketCow.getBucketInfo().getName(), randomName());
+    createBlobWithContents(storageCow, blobId, "blob-contents");
```
wchamber, line 101: do we need to give the blob any contents, or can we just create the blob?

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -86,10 +94,14 @@ public void tearDownPubsub() throws Exception {
   @Test
   public void subscribeAndCleanupResource_googleBucket() throws Exception {
```
wchamber, line 95: per offline talk, let's have a test that checks that an already deleted bucket ends up in the DONE TrackedResourceState. Probably another PR.

yonghaoy: Added

PR #46, opened 2020-08-20
[PF-32]: Support delete Google Blob

yonghaoy:


src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBlobCleanupFlight.java
```diff
@@ -0,0 +1,25 @@
...
+
+import bio.terra.cloudres.common.ClientConfig;
+import bio.terra.janitor.db.JanitorDao;
+import bio.terra.stairway.Flight;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.RetryRuleFixedInterval;
+import org.springframework.context.ApplicationContext;
+
+/** A Flight to cleanup Blob in GCP. */
+public class GoogleBlobCleanupFlight extends Flight {
```
yonghaoy, line 11: @wchamber I can also make an abstract ResourceCleanupFlight and all resource externs that class. What do you prefer?

wchamber: Hmmm. I'm not sure we get a lot from having a base flight class. I guess it would handle the initial/cleanup steps, some default retry, but that's not a lot. WDYT?
I'm ok either way, and we can always change our mind later.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBlobCleanupStep.java
```diff
@@ -0,0 +1,43 @@
...
+  private Logger logger = LoggerFactory.getLogger(GoogleBlobCleanupStep.class);
+
+  public GoogleBlobCleanupStep(ClientConfig clientConfig, JanitorDao janitorDao) {
+    super(clientConfig, janitorDao);
+  }
+
+  @Override
+  protected StepResult cleanUp(CloudResourceUid resourceUid) {
+    try {
+      // Delete all Blobs before deleting bucket.
```
wchamber, line 25: nit: left over copy paste

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBlobCleanupStep.java
```diff
@@ -0,0 +1,43 @@
...
+      BlobId blobId =
+          BlobId.of(
+              resourceUid.getGoogleBlobUid().getBucketName(),
+              resourceUid.getGoogleBlobUid().getBlobName());
+      BlobCow blobCow = storageCow.get(blobId);
+      if (blobCow == null) {
+        // Blob doesn't exists.
+        return StepResult.getStepResultSuccess();
+      }
+      blobCow.delete();
```
wchamber, line 35: what happens if we skip the `get` and just try to delete the blob directly?

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -108,6 +108,28 @@ public void subscribeAndCleanupResource_googleBucket() throws Exception {
     assertNull(storageCow.get(blobId));
   }
 
+  @Test
+  public void subscribeAndCleanupResource_googleBlob() throws Exception {
```
wchamber, line 112: probably should also have a test for deleting a blob that's already been deleted, or when the bucket's already deleted. WDYT?

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleBlobCleanupStep.java
yonghaoy: Ah, right, remove the extra get

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
yonghaoy: added a new one

PR #48, opened 2020-08-26
[PF-9] Add admin user authZ support

yonghaoy:
The current implementation uses a config which contains list of admin user email address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,3 +106,18 @@ Stop the local postgres:
 ```
 local-dev/run_postgres.sh stop
 ```
+## Authorization
+Janitor uses a list of user email address as administrator user to access its admin endpoint.
+The value can be set in `iam.adminUserList` configuration.
+### For Broad Engineers:
+The file is stored in Vault. 
+To read user list, run:
+```
+docker run -e VAULT_TOKEN=$(cat ~/.vault-token) -it broadinstitute/dsde-toolbox:dev vault read config/terra/crl-janitor/common/iam 
+```
+To request admin access, please contact [mc-terra-janitor-admins](https://github.com/orgs/broadinstitute/teams/mc-terra-janitor-admins) members.
+
+To update user list, run:
+```
+docker run --rm --cap-add IPC_LOCK -e "VAULT_TOKEN=$(cat ~/.vault-token)" -e "VAULT_ADDR=https://clotho.broadinstitute.org:8200" -v $(pwd):/current vault:1.1.0 vault write config/terra/crl-janitor/common/iam admin-users=@/current/{USER_FILE}}
+```
--- a/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
@@ -18,13 +18,15 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
 public class IamConfiguration {
   private static final Logger logger = LoggerFactory.getLogger(IamConfiguration.class);
 
-  // The file path for admin user list.
+  // The Janitor admin user list.
   private String adminUserList;
+  // Extracted from the adminUserList value. Mutually exclusive with adminUserList.
   private Set<String> adminUsers;
-  private boolean configBasedAuthZEnabled;
+  // If the config based authorization are enabled. If disabled, Janitor will not perform the config
+  // authZ check.
+  private boolean configBasedAuthzEnabled;
 
   public Set<String> getAdminUsers() {
-    logger.warn("Failed to read admin user file from configuration" + adminUserList);
     return adminUsers;
   }
 
@@ -42,17 +44,16 @@ public class IamConfiguration {
                   adminUserList,
                   TypeFactory.defaultInstance().constructCollectionType(Set.class, String.class));
     } catch (IOException e) {
-      logger.warn("Failed to read admin user file from configuration", e);
-      throw new RuntimeException(
-          "Failed to read admin user file from configuration", e);
+      logger.warn("Failed to read admin user list from configuration", e);
+      throw new RuntimeException("Failed to read admin user list from configuration", e);
     }
   }
 
-  public boolean isConfigBasedAuthZEnabled() {
-    return configBasedAuthZEnabled;
+  public boolean isConfigBasedAuthzEnabled() {
+    return configBasedAuthzEnabled;
   }
 
-  public void setConfigBasedAuthZEnabled(boolean configBasedAuthZEnabled) {
-    this.configBasedAuthZEnabled = configBasedAuthZEnabled;
+  public void setConfigBasedAuthzEnabled(boolean configBasedAuthzEnabled) {
+    this.configBasedAuthzEnabled = configBasedAuthzEnabled;
   }
 }
--- a/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
@@ -1,0 +1,56 @@
+package bio.terra.janitor.app.configuration;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.type.TypeFactory;
+import java.io.IOException;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+@Component
+@EnableConfigurationProperties
+@EnableTransactionManagement
+@ConfigurationProperties(prefix = "iam")
+public class IamConfiguration {
+  private static final Logger logger = LoggerFactory.getLogger(IamConfiguration.class);
+
+  // The file path for admin user list.
+  private String adminUserList;
+  private Set<String> adminUsers;
+  private boolean configBasedAuthZEnabled;
+
+  public Set<String> getAdminUsers() {
+    return adminUsers;
+  }
+
+  public void setAdminUserList(String adminUserList) {
+    this.adminUserList = adminUserList;
+    parseAdminUserFromConfig();
+  }
+
+  /** Reads {@code adminUserList} and convert to a {@code Set} */
+  private void parseAdminUserFromConfig() {
+    try {
+      adminUsers =
+          new ObjectMapper()
+              .readValue(
+                  adminUserList,
+                  TypeFactory.defaultInstance().constructCollectionType(Set.class, String.class));
+    } catch (IOException e) {
+      logger.warn("Failed to read admin user file from configuration", e);
+      throw new RuntimeException("Failed to read admin user file from configuration", e);
+    }
+  }
+
+  public boolean isConfigBasedAuthZEnabled() {
+    return configBasedAuthZEnabled;
+  }
+
+  public void setConfigBasedAuthZEnabled(boolean configBasedAuthZEnabled) {
+    this.configBasedAuthZEnabled = configBasedAuthZEnabled;
+  }
+}
--- a/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
@@ -24,6 +24,7 @@ public class IamConfiguration {
   private boolean configBasedAuthZEnabled;
 
   public Set<String> getAdminUsers() {
+    logger.warn("Failed to read admin user file from configuration" + adminUserList);
     return adminUsers;
   }
 
@@ -41,8 +42,9 @@ public class IamConfiguration {
                   adminUserList,
                   TypeFactory.defaultInstance().constructCollectionType(Set.class, String.class));
     } catch (IOException e) {
-      logger.warn("Failed to read admin user file from configuration", e);
-      throw new RuntimeException("Failed to read admin user file from configuration", e);
+      logger.warn("Failed to read admin user file from configuration" + adminUserList, e);
+      throw new RuntimeException(
+          "Failed to read admin user file from configuration" + adminUserList, e);
     }
   }
 
--- a/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
@@ -42,9 +42,9 @@ public class IamConfiguration {
                   adminUserList,
                   TypeFactory.defaultInstance().constructCollectionType(Set.class, String.class));
     } catch (IOException e) {
-      logger.warn("Failed to read admin user file from configuration" + adminUserList, e);
+      logger.warn("Failed to read admin user file from configuration", e);
       throw new RuntimeException(
-          "Failed to read admin user file from configuration" + adminUserList, e);
+          "Failed to read admin user file from configuration", e);
     }
   }
 
--- a/src/main/java/bio/terra/janitor/app/controller/JanitorApiController.java
+++ b/src/main/java/bio/terra/janitor/app/controller/JanitorApiController.java
@@ -2,6 +2,8 @@ package bio.terra.janitor.app.controller;
 
 import bio.terra.generated.controller.JanitorApi;
 import bio.terra.generated.model.*;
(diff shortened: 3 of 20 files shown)


src/main/java/bio/terra/janitor/service/iam/AuthenticatedUserRequest.java
```diff
@@ -0,0 +1,67 @@
...
+  private String email;
+  private String subjectId;
+  private Optional<String> token;
+  private UUID reqId;
+
+  public AuthenticatedUserRequest() {
+    this.reqId = UUID.randomUUID();
+  }
+
+  public AuthenticatedUserRequest(String email, String subjectId, Optional<String> token) {
```
gjuggler, line 18: Taking an optional as constructor / method param is a bit questionable (the most convincing argument I've seen is something like https://stackoverflow.com/a/39005452). I also can't find any usage of the token field in this PR. Following go/yagni principle, maybe scrap the token field altogether for now?

src/main/java/bio/terra/janitor/service/iam/IamService.java
```diff
@@ -0,0 +1,33 @@
...
+public class IamService {
+  private final IamConfiguration iamConfiguration;
+
+  @Autowired
+  public IamService(IamConfiguration iamConfiguration) {
+    this.iamConfiguration = iamConfiguration;
+  }
+
+  /** Check if user is an administrator. Throws {@link UnauthorizedException} if not. */
+  public void isAdminUser(AuthenticatedUserRequest userReq) {
```
gjuggler, line 24: nit: generally a method starting with `is` would expect a boolean return. Maybe rename this to `requireAdminUser`?

src/main/java/bio/terra/janitor/service/iam/IamService.java
```diff
@@ -0,0 +1,33 @@
...
+    if (iamConfiguration.isConfigBasedAuthZEnabled()) {
```
gjuggler, line 25: This doesn't feel right – we want to return early if config-based authz is *not* enabled, correct?

src/main/java/bio/terra/janitor/service/iam/IamService.java
```diff
@@ -0,0 +1,33 @@
...
+      return;
+    }
+    boolean isAdmin = iamConfiguration.getAdminUsers().contains(userReq.getEmail());
+    if (!isAdmin)
```
gjuggler, line 29: style nit: has there already been a discussion about adhering to Google's Java style guide? They prefer braces everywhere https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

(That said, my attitude is that we generally shouldn't worry at all about stuff like this, and we should aim to add an auto-formatter sooner than later to make things reliably consistent w/ a known style guide. AoU used Spotless, which worked really well for Java code. I can set that up for MCT repos.)

src/main/java/bio/terra/janitor/service/iam/ProxiedAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,30 @@
+package bio.terra.janitor.service.iam;
+
+import java.util.Optional;
+import javax.servlet.http.HttpServletRequest;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Component;
+
+/**
+ * Implementation of {link AuthenticatedUserRequestFactory} when HTTP requests enter from Apache
+ * Proxy. In this scenario, Janitor service is deployed behind Apache Proxy, and Apache Proxy
```
gjuggler, line 10: This comment feels like it's truncated.

It may be worth noting why we can trust these headers to not be abused, e.g. "Apache proxy clears out any inbound values for these headers, so they are guaranteed to contain valid auth information."

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
```diff
@@ -22,14 +24,24 @@
...
 
-  public CreatedResource createResource(CreateResourceRequestBody body) {
+  public CreatedResource createResource(
+      CreateResourceRequestBody body, AuthenticatedUserRequest userReq) {
+    iamService.isAdminUser(userReq);
+    return createResourceInternal(body);
+  }
+
+  /** Internal method to Create a new {@link TrackedResource} without user credentials. */
+  public CreatedResource createResourceInternal(CreateResourceRequestBody body) {
```
gjuggler, line 44: nit: no longer public?

src/main/java/bio/terra/janitor/service/iam/IamService.java
```diff
@@ -0,0 +1,33 @@
+package bio.terra.janitor.service.iam;
+
+import bio.terra.janitor.app.configuration.IamConfiguration;
+import bio.terra.janitor.common.exception.UnauthorizedException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * IAM Service which handles authorization. Currently, Janitor uses a simple user email allow list
+ * to check user permission.
```
gjuggler, line 10: Please add somewhere in this class-level comment (and maybe somewhere in the README too) some information about how service-level auth works here. My understanding (I had to go back & read the Janitor auth doc to refresh my memory) is that all Janitor clients will make Janitor calls with an access token from a specific service account, and this service account will be included in the allowlist?

src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+public class IamConfiguration {
+  private static final Logger logger = LoggerFactory.getLogger(IamConfiguration.class);
+
+  // The file path for admin user list.
+  private String adminUserList;
+  private Set<String> adminUsers;
+  private boolean configBasedAuthZEnabled;
+
+  public Set<String> getAdminUsers() {
+    logger.warn("Failed to read admin user file from configuration" + adminUserList);
```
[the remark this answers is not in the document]
yonghaoy: ah, sorry, what is the leftover when I debugged this feature. Deleted

src/main/resources/application.properties
```diff
@@ -19,3 +19,6 @@ stairway.terminateTimeout=5s
 pubsub.track-resource.enabled=${TRACK_RESOURCE_PUBSUB_ENABLED}
 pubsub.track-resource.projectId=${TRACK_RESOURCE_PUBSUB_PROJECT_ID}
 pubsub.track-resource.subscription=${TRACK_RESOURCE_PUBSUB_SUBSCRIPTION}
+# TODO(PF-81): Switch to use SAM instead of config file for user authZ.
+iam.configBasedAuthZEnabled=${CONFIG_BASED_AUTHZ_ENABLED}
+iam.adminUserListFilePath=${ADMIN_USER_LIST}
```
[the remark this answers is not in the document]
yonghaoy: Done

src/main/java/bio/terra/janitor/service/iam/IamService.java
yonghaoy: Actually there is no service authZ for admin endpoints use cases. Those 'admin endpoints are for users'(like admin users run things locally or via scripts). We don't have any use cases in which some services want to access admin endpoints using their service token. 
All services just use the 'specific service account' to publish messages, and service authZ is controlled by pubsub IAM policy. 
So I think the allowlist would just contains user emails. And even if any services in the future want to use admin endpoints, it should trigged by a user,  admin user token is expected in the request header, not service token.

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
yonghaoy: unfortunately it need to be public because classes in other package would need this. (TrackedResourceSubscriber)

README.md
```diff
@@ -106,3 +106,18 @@ Stop the local postgres:
...
+## Authorization
+Janitor uses a list of user email address as administrator user to access its admin endpoint.
+The value can be set in `iam.adminUserList` configuration.
+### For Broad Engineers:
+The file is stored in Vault. 
+To read user list, run:
+```
+docker run -e VAULT_TOKEN=$(cat ~/.vault-token) -it broadinstitute/dsde-toolbox:dev vault read config/terra/crl-janitor/common/iam 
+```
+To request admin access, please contact [mc-terra-janitor-admins](https://github.com/orgs/broadinstitute/teams/mc-terra-janitor-admins) members.
```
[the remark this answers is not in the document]
yonghaoy: The admin list has permission to modify the Vault value. So this section's header is "for broad engineer", because this is where Broad Janitor stores the config value. Other Janitor can have there own way to store the config value, no functional impact

src/main/java/bio/terra/janitor/service/iam/ProxiedAuthenticatedUserRequestFactory.java
yonghaoy: Thanks!

src/main/java/bio/terra/janitor/service/iam/IamService.java
yonghaoy: Some of my intellij auto-complete not adding braces, just manually added it back. 
Janitor also uses Spotless, but I don't think it can add braces like that. 
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/build.gradle#L80

src/main/java/bio/terra/janitor/app/configuration/IamConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+@EnableConfigurationProperties
+@EnableTransactionManagement
+@ConfigurationProperties(prefix = "iam")
+public class IamConfiguration {
+  private static final Logger logger = LoggerFactory.getLogger(IamConfiguration.class);
+
+  // The file path for admin user list.
+  private String adminUserList;
+  private Set<String> adminUsers;
+  private boolean configBasedAuthZEnabled;
```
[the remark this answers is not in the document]
yonghaoy: Added some comments. Thanks.
And I would expect this property to long-lived because we want to disable the authZ check for local server(I changed the value in `run_local.sh`).
And by having this script, it would help for future migration to SAM?

src/main/java/bio/terra/janitor/app/controller/JanitorApiController.java
```diff
@@ -18,16 +20,26 @@
...
+  public JanitorApiController(
+      JanitorService janitorService,
+      HttpServletRequest request,
+      AuthenticatedUserRequestFactory authenticatedUserRequestFactory) {
     this.janitorService = janitorService;
     this.request = request;
+    this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
+  }
+
+  private AuthenticatedUserRequest getAuthenticatedInfo() {
```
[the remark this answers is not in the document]
yonghaoy: Good suggestion! Also, unrelated points, I copied the whole `AuthenticatedUserRequest`, `ProxiedAuthenticatedUserRequestFactory` stuff form WM and TDR(they all use `getAuthenticatedInfo`). And I am hopping there were a central library...

src/main/java/bio/terra/janitor/common/exception/ApiException.java
```diff
@@ -0,0 +1,12 @@
+package bio.terra.janitor.common.exception;
+
+/** API Exception caused by internal server error. */
```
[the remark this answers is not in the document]
yonghaoy: The whole Auth pattern is coppied from WM and TDR:
https://github.com/DataBiosphere/terra-workspace-manager/blob/7343dad1fdcf7e951d3611bfc3240e56edfaec2f/src/main/java/bio/terra/workspace/common/exception/ApiException.java
I think we want to distinguish different types of InternalServerErrorExceptions? Whether it from Api or some other internal workflow?

src/main/java/bio/terra/janitor/common/exception/InternalServerErrorException.java
```diff
@@ -0,0 +1,12 @@
+package bio.terra.janitor.common.exception;
+
+/** This exception maps to HttpStatus.INTERNAL_SERVER_ERROR in the GlobalExceptionHandler. */
```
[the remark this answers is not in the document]
yonghaoy: Actually this is the 1st internal error Janitor has for any http endpoints. And same as above, the whole pattern is the same as TDR and WM, https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/common/exception/InternalServerErrorException.java
https://github.com/DataBiosphere/terra-workspace-manager/blob/7343dad1fdcf7e951d3611bfc3240e56edfaec2f/src/main/java/bio/terra/workspace/common/exception/InternalServerErrorException.java

We have a TODO here:
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/java/bio/terra/janitor/app/controller/GlobalExceptionHandler.java#L20
And I am expecting to keep similar patterns as WM and TDR, then it is easier for us to migrate to the common library?

src/main/java/bio/terra/janitor/common/exception/UnauthorizedException.java
```diff
@@ -0,0 +1,29 @@
+package bio.terra.janitor.common.exception;
+
+import java.util.List;
+import org.springframework.http.HttpStatus;
+
+/** Exception caused by authorization error. */
+public class UnauthorizedException extends ErrorReportException {
+  private static final HttpStatus thisStatus = HttpStatus.UNAUTHORIZED;
+
+  public UnauthorizedException(String message) {
```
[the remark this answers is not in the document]
yonghaoy: Ah, right, only the 1st one is used in Janitor's case.

src/main/java/bio/terra/janitor/service/iam/AuthenticatedUserRequest.java
```diff
@@ -0,0 +1,67 @@
+package bio.terra.janitor.service.iam;
+
+import bio.terra.janitor.common.exception.ApiException;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import java.util.Optional;
+import java.util.UUID;
+
+public class AuthenticatedUserRequest {
```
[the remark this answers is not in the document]
yonghaoy: Thanks. Added a short comment as you suggested.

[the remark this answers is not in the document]
yonghaoy: Janitor uses autovalues a lot. I uses a normal class is to follow WM and TDR's convension. Shall I change to AutoValue? One problem in aothValue I recently we saw is  AutoValue works well with Gson, but for Jackson, it needs some extra setup. Having getter and setter as regular class like that makes Jackson easier to serialize a class.(I am not sure if we need to serialize this class through). 
CRL uses Gson. But Janitor uses Jackson because that is the what Spring's Openapi generator and Stairway uses. 
Not sure what is the common conversion when choosing java libraries for MCTerra? Personally I likes Gson because it is simpler and get ride of the annoying checked exception - `JsonProcessingException`

src/main/java/bio/terra/janitor/service/iam/AuthenticatedUserRequest.java
yonghaoy: Right. Done. Sorry for so many confusions in this classes. When I copied the whole thing from WM, I didn't pay too much attention to refactor this class. Seems WM and TDR has more complicated use cases as they use SAM. Deleted the token for now.

yonghaoy:
Thanks @gjuggler All comments are addressed. PTAL, thanks!

src/main/java/bio/terra/janitor/service/iam/AuthenticatedUserRequest.java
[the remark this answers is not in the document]
gjuggler: Ah, interesting tradeoffs here. IMO it hinges on whether we expect this class to be shared w/ other services in a common library. If so, we should probably be as plain as possible. But if not, then using autovalue sounds reasonable.

Re: the GSON vs. Jackson question, the swagger & stairway dependencies on Jackson probably make it overall the right decision to stick with Jackson everywhere. Hopefully not too much of a drag :) and I'll make sure to add some note on the tradeoffs involved into our tech stack documentation.

src/main/java/bio/terra/janitor/service/iam/IamService.java
gjuggler: Aha, thanks – I had forgotten that we'd decided that service clients won't use HTTP directly.

PR #49, opened 2020-08-27
Use terra-helmfile's new version update workflow

choover-broad:
We have a new, more robust workflow for updating service versions in terra-helmfile (uses yq instead of regex matching). This PR configures janitor to use it.


choover-broad:
(Related: https://github.com/broadinstitute/terra-helmfile/pull/128)

PR #52, opened 2020-08-28
[PF-12] [PF-16] Support abandon and bump resource

yonghaoy:
Abandon Resource: change Resource state from READY or CLEANING to ABANDONED
Bump Resource: change Resource state from ABANDONED to READY

### DIFF
--- a/src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
+++ b/src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
@@ -1,16 +1,17 @@
 package bio.terra.janitor.service.janitor;
 
 import bio.terra.generated.model.*;
+import bio.terra.janitor.common.NotFoundException;
+import bio.terra.janitor.common.exception.InternalServerErrorException;
 import bio.terra.janitor.db.*;
 import bio.terra.janitor.service.iam.AuthenticatedUserRequest;
 import bio.terra.janitor.service.iam.IamService;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
-import java.util.List;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
```diff
@@ -126,11 +127,72 @@ public TrackedResourceInfoList getResources(
...
+          getResourceWithState(cloudResourceUid, TrackedResourceState.ABANDONED);
+
+      // It is possible that there might be multiple ABANDON resources, and we only need to update
+      // the one with last
+      // expiration time.
+      TrackedResource latestResource =
+          resources.stream().max(Comparator.comparing(TrackedResource::expiration)).get();
+      janitorDao.updateResourceState(
+          latestResource.trackedResourceId(), TrackedResourceState.READY);
+    }
```
gjuggler, line 172: If we add additional ResourceState enum values in the future, it would be easy to forget to add another conditional block here. Maybe add a catch-all else statement which returns an internal server error?

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -219,4 +218,204 @@ public void getResources() throws Exception {
             .collect(Collectors.toList()),
         Matchers.containsInAnyOrder(id1, id2));
   }
+
+  @Test
+  public void abandonThenBumpResources() throws Exception {
```
gjuggler, line 223: [Side note / comment] Not directly related to this PR, but I feel like it's not quite right to be testing the service business logic from the controller level. It adds a fair amount of boilerplate, since we're dealing with things like serializing requests and tracking URL paths. I'm curious if there's been much discussion within or across teams about the aesthetic of what layer to run Spring-based unit tests. (We've had many such discussions on the AoU team, and I'm happy to share our experiences.)

gjuggler, line 223: [optional] I feel like "abandon resource" and "bump resource" are two well-separated pieces of functionality that are better off tested independently. This test is >100 lines long... maybe consider splitting it up into an abandonResource and then a bumpResource test (even if you have to share some setup code)?

yonghaoy: Done

src/main/java/bio/terra/janitor/service/janitor/JanitorService.java
yonghaoy: Done!

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
yonghaoy: We haven't discussed the level of test. The 1st test we added is createResource test. 
At that time, I am not confident about the open api schema, so I want the test starts from controller then I can verify the url, query, response code all together.  
Seems WM doesn't have controller test but they have service test.
While Data Repo has controller test: https://github.com/DataBiosphere/jade-data-repo/tree/develop/src/test/java/bio/terra/app/controller and that is where I learnt the test structure.
Happy to chat more about the test structure. Thanks!

PR #54, opened 2020-09-03
[PF-16] Refactor bump/abandon test and add TrackedResourceService 

yonghaoy:
Also, I adressed two comments from previous PR but forgot to push them to master:
https://github.com/DataBiosphere/terra-resource-janitor/pull/52#discussion_r479413942
And https://github.com/DataBiosphere/terra-resource-janitor/pull/52#discussion_r479408656
So this PR also includes the leftover commit which should be in that PR


src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -341,6 +341,121 @@ public void abandonThenBumpResources() throws Exception {
     assertEquals(TrackedResourceState.READY.name(), newerResource.getState());
   }
 
+  @Test
+  public void bumpResources() throws Exception {
```
wchamber, line 345: I think we have been and and are getting a little worse on what should be a `JanitorServiceTest` vs a `JanitorApiControllerTest`.  Business logic tests like "what does bump do" should be `JanitorServiceTest`. The only thing we need to test as a `JanitorApiControllerTest` is serialization - is the request/response format as expected? Arguably, those are often not very valuable tests once we have a couple examples - it's mostly handled by our libraries; if one serialization works, they probably all work (as long as they're "the same").

So, I think this test is misplaced in that it's doing too much about how create/abandon/bump interact. OTOH, we're maybe doing a little too much of that already in this file, and we do need to set up a little bit in order for "bump" to succeed at all. I'd advocate for making this a test in `JanitorServiceTest` instead, but let me know if you disagree.

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -220,7 +220,7 @@ public void getResources() throws Exception {
   }
 
   @Test
-  public void abandonThenBumpResources() throws Exception {
+  public void abandoResources() throws Exception {
```
wchamber, line 223: What's the difference between this test and `bumpResources` test? Also, see other comment.

wchamber (changes requested):
This should really be two PRs, but i'm not going to block the review on it. The majority of the line changes are unrelated to dropping the cleanup log table.
At the least, update the PR description to give equal weight to these unrelated changes. Next time, please split the changes up.

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
yonghaoy: Yeah, I also realized that those tests should be done in `JanitorServiceTest`, I thought getting a chance to check with you then refactor those tests. But for now in this PR, I can move abandAndBump to  `JanitorServiceTest`, PTAL, thanks.

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
yonghaoy: Done

src/test/java/bio/terra/janitor/service/janitor/JanitorServiceTest.java
```diff
@@ -87,4 +90,67 @@ public void createResource_Duplicates() {
...
+        janitorService
+            .createResource(
+                new CreateResourceRequestBody()
+                    .resourceUid(resourceUid)
+                    .creation(DEFAULT_TIME)
+                    .expiration(DEFAULT_TIME),
+                ADMIN_USER)
+            .getId();
+
+    String secondId =
```
wchamber, line 109: I think we have a second id to test that a duplicated resource is not bumped, which is important business logic so that we don't have duplicate ready resources. add a comment somewhere.

src/test/java/bio/terra/janitor/service/janitor/JanitorServiceTest.java
```diff
@@ -87,4 +90,67 @@ public void createResource_Duplicates() {
...
+  public void abandonResources_notFound() throws Exception {
+    CloudResourceUid resourceUid =
+        new CloudResourceUid()
+            .googleProjectUid(new GoogleProjectUid().projectId(UUID.randomUUID().toString()));
+    assertThrows(
+        NotFoundException.class,
+        () -> janitorService.updateResource(resourceUid, ResourceState.ABANDONED, ADMIN_USER));
+  }
+
+  /** Gets 404 error when resource exists but not in ABANDONED state. */
```
wchamber, line 146: there's no 404 in this test.

src/test/java/bio/terra/janitor/service/janitor/JanitorServiceTest.java
```diff
@@ -87,4 +90,67 @@ public void createResource_Duplicates() {
...
+  }
+
+  @Test
+  public void abandonResources_notFound() throws Exception {
+    CloudResourceUid resourceUid =
+        new CloudResourceUid()
+            .googleProjectUid(new GoogleProjectUid().projectId(UUID.randomUUID().toString()));
+    assertThrows(
+        NotFoundException.class,
+        () -> janitorService.updateResource(resourceUid, ResourceState.ABANDONED, ADMIN_USER));
```
wchamber, line 143: Now that I'm looking at the implementation and these tests, I have a few issues with `JanitorService#updateResource`

1. I don't think updateResource should be a single method that only does something for READY and ABANDONED. `updateResource` sounds like any tracked resource should be allowed. What we actually want to allow is 2 very specific state transitions, bumping and abandoning. Those both have resource states that they make sense for, and some that they don't. They're logically different and should be so. We should also differentiate in the response, I think, between 1) there's no tracked resource with this resourceUid and 2) there's no tracked resource with this resourceUid in a bump-able state.
2. I don't think the JanitorService should be doing authorization. That's a higher level concern than the JanitorService's "how does a resource get bumped or abandoned". A good signal that this is out of place is that the other JanitorService method has two public methods for creating resources, one with auth and one without that's "Internal." The api controller does care that requests are authorized to do different logical actions, but we don't need to propogate that down to the next layer of how they are executed.

wchamber (changes requested):

src/test/java/bio/terra/janitor/service/janitor/JanitorServiceTest.java
yonghaoy: created one more layer as synced offline

src/main/java/bio/terra/janitor/service/janitor/JanitorApiService.java
```diff
@@ -0,0 +1,73 @@
...
+import bio.terra.janitor.service.iam.IamService;
+import java.util.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.support.TransactionTemplate;
+
+@Component
+public class JanitorApiService {
+  private final JanitorDao janitorDao;
+  private final TransactionTemplate transactionTemplate;
```
wchamber, line 16: nit: i don't think this class needs the TransactionTemplate anymore.

src/main/java/bio/terra/janitor/service/janitor/JanitorApiService.java
```diff
@@ -0,0 +1,73 @@
...
+import bio.terra.janitor.db.*;
+import bio.terra.janitor.service.iam.AuthenticatedUserRequest;
+import bio.terra.janitor.service.iam.IamService;
+import java.util.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.support.TransactionTemplate;
+
+@Component
+public class JanitorApiService {
```
wchamber, line 14: add a comment about what this service is for

src/main/java/bio/terra/janitor/service/janitor/JanitorApiService.java
```diff
@@ -0,0 +1,73 @@
...
+    this.trackedResourceService = trackedResourceService;
+  }
+
+  public CreatedResource createResource(
+      CreateResourceRequestBody body, AuthenticatedUserRequest userReq) {
+    iamService.requireAdminUser(userReq);
+    return trackedResourceService.createResource(body);
+  }
+
+  /** Retrieves the info about a tracked resource if their exists a resource for that id. */
```
wchamber, line 38: nit: `there` instead of `their` (was probably my mistake initially)

src/main/java/bio/terra/janitor/service/janitor/JanitorApiService.java
```diff
@@ -0,0 +1,73 @@
...
+  public void updateResource(
+      CloudResourceUid cloudResourceUid, ResourceState state, AuthenticatedUserRequest userReq) {
+    iamService.requireAdminUser(userReq);
+    if (state == ResourceState.ABANDONED) {
+      trackedResourceService.abandonResource(cloudResourceUid);
+    } else if (state == ResourceState.READY) {
+      trackedResourceService.bumpResource(cloudResourceUid);
+    } else {
+      throw new InternalServerErrorException(
+          String.format("Unhandled state: %s for resource %s", state, cloudResourceUid));
```
wchamber, line 70: this should throw some bad request exception. We don't need to include the cloudResourceUid, just the state.

src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
```diff
@@ -116,60 +108,47 @@ private TrackedResourceId createResourceAndUpdateDuplicates(
...
-   * Updates the resource state.
-   *
-   * <p>Currently it supports:
+   * Updates the ABANDONED resource state to READY.
    *
-   * <ul>
-   *   <li>Abandon resource: Update resource from READY or CLEANING to ABANDONED
-   *   <li>Bump resource: Update resource from ABANDONED to READY
-   * </ul>
+   * <p>It is possible that there might be multiple ABANDON resources, and we only need to update
```
wchamber, line 142: nit: `multiple ABANDONED`

src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
```diff
@@ -116,60 +108,47 @@ private TrackedResourceId createResourceAndUpdateDuplicates(
...
-      }
-      janitorDao.updateResourceState(
-          resources.get(0).trackedResourceId(), TrackedResourceState.ABANDONED);
-    } else if (state == ResourceState.READY) {
-      // Bump ABANDONED state resources for cleaning.
-      List<TrackedResource> resources =
-          getResourceWithState(cloudResourceUid, TrackedResourceState.ABANDONED);
+  public void bumpResource(CloudResourceUid cloudResourceUid) {
+    List<TrackedResource> resources =
+        getResourceWithState(cloudResourceUid, TrackedResourceState.ABANDONED);
```
wchamber, line 147: We should also be able to bump ERROR resources, not just ABANDONED. also update comment.

src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
```diff
@@ -116,60 +108,47 @@ private TrackedResourceId createResourceAndUpdateDuplicates(
...
 
+  /** Updates the READY resource state to ABANDONED. */
+  public void abandonResource(CloudResourceUid cloudResourceUid) {
+    List<TrackedResource> resources =
+        getResourceWithState(
+            cloudResourceUid, TrackedResourceState.READY, TrackedResourceState.CLEANING);
+
+    if (resources.size() > 1) {
+      throw new InternalServerErrorException(
+          "More than one READY or CLEANING state resources are found.");
```
wchamber, line 133: Hmmm, maybe we should log an error but still abandon all of the resources anyway. If we're in a bad state where there are multiple active resources, abandoning all of them seems reasonable. We do want to indicate with a log error that it's a bad state. WDYT?

src/test/java/bio/terra/janitor/service/janitor/JanitorApiServiceTest.java
```diff
@@ -52,39 +55,104 @@ public void createResource_Duplicates() {
...
-    retrievedStates = extractStates(janitorService.getResources(resourceUid, ADMIN_USER));
+    retrievedStates = extractStates(janitorApiService.getResources(resourceUid, ADMIN_USER));
     assertThat(retrievedStates, Matchers.hasEntry(firstId, "DUPLICATED"));
     assertThat(retrievedStates, Matchers.hasEntry(secondId, "DUPLICATED"));
     assertThat(retrievedStates, Matchers.hasEntry(thirdId, "READY"));
     assertThat(retrievedStates, Matchers.aMapWithSize(3));
   }
+
+  @Test
+  public void abandonThenBumpResources() throws Exception {
```
wchamber, line 95: I think these tests should now move from the JanitorApiServiceTEst to the `TrackedResourceServiceTest`. That's where the logic is.
The JanitorApiServiceTest might not have many if any tets anymore. The only logic in it to be tested is checking that the request is authenticated, which is arguably not worth writing tests for since it's a one liner.

(other tests in this file as well)

src/test/java/bio/terra/janitor/service/pubsub/TrackedResourceSubscriberTest.java
```diff
@@ -42,7 +42,7 @@
   @Qualifier(OBJECT_MAPPER)
   private ObjectMapper objectMapper;
 
-  @Autowired private JanitorService janitorService;
+  @Autowired private TrackedResourceService trackedResourceService;
 
   @Autowired private MockMvc mvc;
```
wchamber, line 47: I don't think this test should have an authenticated user or MockMvc.

wchamber (changes requested):

src/main/java/bio/terra/janitor/service/janitor/JanitorApiService.java
yonghaoy: Done

src/test/java/bio/terra/janitor/service/pubsub/TrackedResourceSubscriberTest.java
yonghaoy: Done

src/main/java/bio/terra/janitor/service/janitor/JanitorApiService.java
yonghaoy: Done

src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
yonghaoy: DONE

src/test/java/bio/terra/janitor/service/janitor/JanitorApiServiceTest.java
yonghaoy: Done

src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
yonghaoy: Done

wchamber (approved):

PR #55, opened 2020-09-04
[PF-99] Add more logs in stairway

yonghaoy:
And for abandon/bump resource. I feel that is not right to use trackedResourceId because the request takes CloudResourceUid and TrackedResourceId is loaded in the middle of the request:
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/java/bio/terra/janitor/service/janitor/JanitorService.java#L152

And I don't think we need to add the TrackedResourceId/CloudResourceUid into MDC because the logic is simple, not too many things worth to log. Maybe in the future once we worked on Tracing ticket, we can add requestId into MDC then WM:
https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/app/configuration/ApiResourceConfig.java#L41

### DIFF
--- a/src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
+++ b/src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
@@ -11,18 +11,19 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
-public class StairwayLoggingHooks implements StairwayHook {
-  private static final String FlightLogFormat =
+/** {@link StairwayHook} to add log info for cleanup flight. */
+public class CleanupLoggingHook implements StairwayHook {
+  private static final String FLIGHT_LOG_FORMAT =
       "Operation: {}, flightClass: {}, flightId: {}, timestamp: {}";
   private static final String StepLogFormat =
       "Operation: {}, flightClass: {}, flightId: {}, stepIndex: {}," + "timestamp: {}";
   private static final String TRACKED_RESOURCE_MDC_KEY = "trackedResourceId";
-  private static final Logger logger = LoggerFactory.getLogger(StairwayHook.class);
+  private static final Logger logger = LoggerFactory.getLogger(CleanupLoggingHook.class);
 
   @Override
   public HookAction startFlight(FlightContext context) {
     logger.info(
-        FlightLogFormat,
+        FLIGHT_LOG_FORMAT,
         "startFlight",
         context.getFlightClassName(),
         context.getFlightId(),
@@ -52,7 +53,7 @@ public class StairwayLoggingHooks implements StairwayHook {
   @Override
   public HookAction endFlight(FlightContext context) {
     logger.info(
-        FlightLogFormat,
+        FLIGHT_LOG_FORMAT,
         "endFlight",
         context.getFlightClassName(),
         context.getFlightId(),
--- a/src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
+++ b/src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
@@ -5,8 +5,6 @@ import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.HookAction;
 import bio.terra.stairway.StairwayHook;
 import java.time.Instant;
-import java.time.ZoneId;
-import java.time.format.DateTimeFormatter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
@@ -27,7 +25,7 @@ public class CleanupLoggingHook implements StairwayHook {
         "startFlight",
         context.getFlightClassName(),
         context.getFlightId(),
-        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
+        Instant.now().toString());
     return HookAction.CONTINUE;
   }
 
@@ -39,7 +37,7 @@ public class CleanupLoggingHook implements StairwayHook {
         context.getFlightClassName(),
         context.getFlightId(),
         context.getStepIndex(),
-        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
+        Instant.now().toString());
     TrackedResourceId trackedResourceId =
         context
             .getInputParameters()
@@ -57,7 +55,7 @@ public class CleanupLoggingHook implements StairwayHook {
         "endFlight",
         context.getFlightClassName(),
         context.getFlightId(),
-        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
+        Instant.now().toString());
     return HookAction.CONTINUE;
   }
 
@@ -69,7 +67,7 @@ public class CleanupLoggingHook implements StairwayHook {
         context.getFlightClassName(),
         context.getFlightId(),
         context.getStepIndex(),
-        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
+        Instant.now().toString());
     return HookAction.CONTINUE;
   }
 }
--- a/src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
+++ b/src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
@@ -4,7 +4,6 @@ import bio.terra.janitor.db.TrackedResourceId;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.HookAction;
 import bio.terra.stairway.StairwayHook;
-import java.time.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
@@ -12,9 +11,9 @@ import org.slf4j.MDC;
 /** {@link StairwayHook} to add log info for cleanup flight. */
 public class CleanupLoggingHook implements StairwayHook {
   private static final String FLIGHT_LOG_FORMAT =
-      "Operation: {}, flightClass: {}, flightId: {}, timestamp: {}";
+      "Operation: {}, flightClass: {}, flightId: {}";
   private static final String StepLogFormat =
-      "Operation: {}, flightClass: {}, flightId: {}, stepIndex: {}," + "timestamp: {}";
+      "Operation: {}, flightClass: {}, flightId: {}, stepIndex: {}";
   private static final String TRACKED_RESOURCE_MDC_KEY = "trackedResourceId";
   private static final Logger logger = LoggerFactory.getLogger(CleanupLoggingHook.class);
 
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
```diff
@@ -17,6 +17,7 @@ public FlightSubmission createSubmission(TrackedResource trackedResource) {
     FlightMap flightMap = new FlightMap();
     flightMap.put(FlightMapKeys.TRACKED_RESOURCE_ID, trackedResource.trackedResourceId());
     flightMap.put(FlightMapKeys.CLOUD_RESOURCE_UID, trackedResource.cloudResourceUid());
+    flightMap.put(FlightMapKeys.MDC_KEY, trackedResource.cloudResourceUid());
```
wchamber, line 20: I think it's a little funny to distinguish between the `CLOUD_RESOURCE_UID` and the `MDC_KEY` at this stage. When do we expect them to be different? Why not just look up the `CLOUD_RESOURCE_UID` in the hook and use that?

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+import bio.terra.stairway.HookAction;
+import bio.terra.stairway.StairwayHook;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+public class StairwayLoggingHooks implements StairwayHook {
```
wchamber, line 14: How about `CleanupLoggingHook`? at least `Hook` instead of `Hooks` to match interface.
Add a javadoc comment for the class.

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+  private static final String FlightLogFormat =
```
wchamber, line 15: nit: `private static final` constant variables should be `CAPITAL_CASE`
ditto others

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+      "Operation: {}, flightClass: {}, flightId: {}, timestamp: {}";
+  private static final String StepLogFormat =
+      "Operation: {}, flightClass: {}, flightId: {}, stepIndex: {}," + "timestamp: {}";
+  private static final String TRACKED_RESOURCE_MDC_KEY = "trackedResourceId";
+  private static final Logger logger = LoggerFactory.getLogger(StairwayHook.class);
+
+  @Override
+  public HookAction startFlight(FlightContext context) {
+    logger.info(
+        FlightLogFormat,
```
wchamber, line 25: why is it important to re-use the same FLIGHT_LOG_FORMAT for startFlight and endFlight?

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+        "startFlight",
+        context.getFlightClassName(),
+        context.getFlightId(),
+        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
```
wchamber, line 29: Logs usually include a timestamp. Why do we need to add our own timestamp? ditto others
If we do, can we just use Instant.now().toString()?

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+        "startStep",
+        context.getFlightClassName(),
+        context.getFlightId(),
+        context.getStepIndex(),
+        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
+    TrackedResourceId trackedResourceId =
+        context
+            .getInputParameters()
+            .get(FlightMapKeys.TRACKED_RESOURCE_ID, TrackedResourceId.class);
+    if (trackedResourceId != null) {
```
wchamber, line 46: when would this not be null? should we log an error?

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+public class StairwayLoggingHooks implements StairwayHook {
+  private static final String FlightLogFormat =
+      "Operation: {}, flightClass: {}, flightId: {}, timestamp: {}";
+  private static final String StepLogFormat =
+      "Operation: {}, flightClass: {}, flightId: {}, stepIndex: {}," + "timestamp: {}";
+  private static final String TRACKED_RESOURCE_MDC_KEY = "trackedResourceId";
+  private static final Logger logger = LoggerFactory.getLogger(StairwayHook.class);
```
wchamber, line 20: why is this using StairwayHook.class instead of its own class?

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public HookAction startStep(FlightContext context) {
+    logger.info(
+        StepLogFormat,
+        "startStep",
+        context.getFlightClassName(),
+        context.getFlightId(),
+        context.getStepIndex(),
+        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
```
wchamber, line 41: Stairway does its own logging that each step starts end. This isn't adding much that we can use since it's not even including the tracked resource id. I'm not sure this is helpful to add, and if we do want it, let's do it after the mdc scope is set.

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public HookAction endStep(FlightContext context) {
+    logger.info(
+        StepLogFormat,
+        "endStep",
+        context.getFlightClassName(),
+        context.getFlightId(),
+        context.getStepIndex(),
+        Instant.now().atZone(ZoneId.of("Z")).format(DateTimeFormatter.ISO_INSTANT));
+    return HookAction.CONTINUE;
```
wchamber, line 72: I would think we have to remove the tracked resource MDC key as well otherwise the context stays on the thread.

wchamber (changes requested):

wchamber:
I'm not convinced that it's not important to log the tracked resource id during bump/abandon. I don't think we should be doing the logging for that at the start of the request as WM does, but I do think we should log a message. That doesn't need to use MDC, but I think it's important to still have a log record of which tracked resource ids are duplicated/bumped/abandoned.
Let's not do it in this PR regardless.

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
yonghaoy: Done

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
yonghaoy: Done

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
yonghaoy: I am not sure I understand this comment. This is to define the common static string variables then we can reuse.

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
yonghaoy: Not every steps has trackedResourceId. We added it here to make sure cleanup related steps gets the trackedResourceId into MDC. 
For example unSupportedFlight, it doesn't have trackedResourceId

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
yonghaoy: This is the pattern I saw in data repo:
https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/job/StairwayLoggingHooks.java
I thought it might be good to add log flight with more context. So I copied the whole thing over

src/main/java/bio/terra/janitor/service/cleanup/StairwayLoggingHooks.java
yonghaoy: Done

src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
yonghaoy: Not needed any more

src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
```diff
@@ -0,0 +1,73 @@
...
+        FLIGHT_LOG_FORMAT,
+        "endFlight",
+        context.getFlightClassName(),
+        context.getFlightId(),
+        Instant.now().toString());
+    return HookAction.CONTINUE;
+  }
+
+  @Override
+  public HookAction endStep(FlightContext context) {
```
wchamber, line 63: this step needs to remove the mdc key so that the thread does not continue to associate the tracked resource id with all of its logs once the step has finished.
`MDC.remove(TRACKED_RESORUCE_MDC_KEY);`

src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(CleanupLoggingHook.class);
+
+  @Override
+  public HookAction startFlight(FlightContext context) {
+    logger.info(
+        FLIGHT_LOG_FORMAT,
+        "startFlight",
+        context.getFlightClassName(),
+        context.getFlightId(),
+        Instant.now().toString());
```
wchamber, line 28: I don't think the timestamp is useful. logs already have timestamps, I know its copied, but let's remove it. ditto others

wchamber (approved):

src/main/java/bio/terra/janitor/service/cleanup/CleanupLoggingHook.java
yonghaoy: DONE

PR #57, opened 2020-09-09
[PF-99]: Log tracked resource id when state changes 

yonghaoy:


src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
```diff
@@ -185,4 +183,10 @@ private static TrackedResourceInfo createInfo(TrackedResourceAndLabels resourceA
         .expiration(OffsetDateTime.ofInstant(resource.expiration(), ZoneOffset.UTC))
         .labels(resourceAndLabels.labels());
   }
+
+  private void abandonResourcesAndLog(TrackedResource trackedResource) {
+    logger.info("Abandoned resource, trackedResourceId: {}", trackedResource.trackedResourceId());
```
wchamber, line 188: only log after the update has succeeded.

src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
```diff
@@ -185,4 +183,10 @@ private static TrackedResourceInfo createInfo(TrackedResourceAndLabels resourceA
         .expiration(OffsetDateTime.ofInstant(resource.expiration(), ZoneOffset.UTC))
         .labels(resourceAndLabels.labels());
   }
+
+  private void abandonResourcesAndLog(TrackedResource trackedResource) {
```
wchamber, line 187: nit: maybe just `abandonResources`. I don't think logging is important enough to include in the function name.

wchamber (approved):

PR #59, opened 2020-09-11
[PF-103] Migrate Janitor app configs to use best practices

gjuggler:
Context: [PF-103](https://broadworkbench.atlassian.net/browse/PF-103) and [Config Management conventions](https://docs.google.com/document/d/1VfgL8D6PpJg7irbL_gKnmyZ5_GH5OYGfK8GPN45odUs/edit#) doc.

This PR attempts to bring Janitor's application config in alignment with the above proposed conventions to the greatest extent possible, without making any breaking changes.

Not directly related to this PR, but one nice side effect of nesting all configs under a 'janitor' prefix is that we now have an obvious hierarchy for overriding configs via env variables, without manually interpolating them within our application config as we do today. And eventually, we can use k8s ConfigMap key-value pairs to set these values too. I quickly validated this functionality by running the below:

```
# Setting this env variable causes unit tests to fail, since the db username is overridden.
JANITOR_DB_USERNAME=asdf ./gradlew unitTest
```


gjuggler:
@wchamber @yonghaoy Thanks for taking a look at the draft on Friday. Since then, I rebased (& incorporated the iam.authz properties) and slightly adjusted the file setup for test properties, to further reduce duplication. I put Yonghao as primary reviewer here.

Integration tests run fine, and the app starts up locally without any issues. I haven't fully tested this on a k8s environment yet – if I should seek some k8s validation before merging, I'll have to bug one of you for some guidance on using kustomize to deploy to my personal cluster.

src/test/resources/application-test.yml
```diff
@@ -0,0 +1,20 @@
+# Config overrides common to all tests should go here.
+# All tests should enable the "test" profile in order to load these property overrides.
```
yonghaoy, line 2: [optonal]: not sure the best practice for naming the file and profile name. Do you think we should name as 'test' or 'testBase', or something else?

yonghaoy (approved):

yonghaoy:
@gjuggler 
Hi Greg, this PR LG! 
And for your question about how to deploy to GKE, we usually use Skaffold. A short description: https://github.com/DataBiosphere/terra-resource-janitor#deploy-to-gke-cluster
But I think I will need to update it a little bit, as we need to cd to local-dev, then do the skaffold things. 
Let me know if you want to deploy your change to GKE but have issues using skaffold

src/test/resources/application-unit.yml
```diff
@@ -0,0 +1,10 @@
+# Config overrides specific to unit tests.
+janitor:
+  primary:
+    # Testing is easier if we are not scheduling changes to the database by default.
+    scheduler-enabled: false
+  pubsub:
+    track-resource:
+      enabled: false
+  stackdriver:
+    enabled: false
```
wchamber, line 10: disabling stackdriver is probably good for the application-test default, thought it's kind of harmless. I think the exporters will fail to set up reasonably.

wchamber (approved):

src/test/resources/application-test.yml
gjuggler: Yeah, I'm not in love with `test` but I think for profile names, brevity is slightly more important than specificity. So I'll stick with `test` for now, we can always adapt later if it's not clear.

src/test/resources/application-unit.yml
gjuggler: Good call – done.

PR #63, opened 2020-09-14
[PF-103] Add a base class for unit & integration tests to further reduce duplication

gjuggler:
I also fixed up some of the Gradle config to show stack traces within failed tests. Not strictly related to the base test class, but it's test-related, so I thought it wasn't worth splitting into a separate cleanup PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+
 buildscript {
     repositories {
         mavenCentral()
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -187,34 +189,40 @@ ideaModule.dependsOn swaggerSources.server.code
...
     ]
-    useJUnitPlatform()
+    testLogging {
+        events = ["passed", "failed", "skipped"]
+        // Causes the correct line to be reported on an exception.
+        exceptionFormat = TestExceptionFormat.FULL
+        reports {
+            // Write XML file (used by CircleCI, Jenkins, etc) to api/build/test-results/test
+            junitXml.enabled = true
+            // Write human-readable test report to api/build/reports/
```
yonghaoy, line 203: what is the api direcotry? I don't think we have the directory yet. How about customize our own report direcroty under build dir? 
I think this may work:
```
 reports.junitXml.destination = file("$buildDir/test-results/test")
 reports.html.destination = file("$buildDir/reports")
```

yonghaoy (approved):

build.gradle
gjuggler: Oops – that was copy-pasta from the AoU repo where the base directory was 'api'. The default directories make sense, and I updated the comment to accurately reflect where they go.

PR #65, opened 2020-09-15
PF-57 Temporarily remove shutdown endpoint.

wchamber:
We're removing the shutdown endpoint which puts the service in a terminating state where it will not run Stairway flights. We do not have an easy way at the moment to restrict the endpoint to k8s lifecycle operations only. Until we do, we're removing the endpoint. See PF-57.

### DIFF
--- a/src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
+++ b/src/main/java/bio/terra/janitor/app/controller/UnauthenticatedApiController.java
@@ -54,21 +54,6 @@ public class UnauthenticatedApiController implements UnauthenticatedApi {
     }
   }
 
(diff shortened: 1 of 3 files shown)


yonghaoy (commented):
I thought you mentioned you will just move this from UnAuthenticated to Admin endpoints? It might be easier to add this back once we have Istio in the future?

wchamber:
> I thought you mentioned you will just move this from UnAuthenticated to Admin endpoints? It might be easier to add this back once we have Istio in the future?

Yeah, I changed my mind. Rather than moving the code around between controllers, adding authz check, and fixing tests for an endpoint we won't use, I opted to remove it entirely. It's also not hard to make a revert commit for this PR.

yonghaoy (approved):

PR #69, opened 2020-09-17
PF-122 Handle FlightNotFound for "completed" cleanup flights

wchamber:
- Handle FlightNotFound while completing flights. Add CleanupFlightState.LOST for this case.
- Add more verbose logging to FlightManager for tracked resource and flight ids.
- Fix config to stop wiping Stairway on start. Keep wiping stairway for tests on start.

### DIFF
--- a/src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
+++ b/src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
@@ -197,15 +197,15 @@ class FlightManager {
    */
   private boolean completeFlight(TrackedResourceAndFlight resourceAndFlight) {
     String flightId = resourceAndFlight.cleanupFlight().flightId();
-    Optional<FlightState> flightState;
+    Optional<FlightStatus> flightStatus;
     try {
-      flightState = Optional.of(stairway.getFlightState(flightId));
+      flightStatus = Optional.of(stairway.getFlightState(flightId).getFlightStatus());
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/janitor/service/cleanup/FlightManager.java
```diff
@@ -197,32 +197,52 @@ public int updateCompletedFlights(int limit) {
    */
   private boolean completeFlight(TrackedResourceAndFlight resourceAndFlight) {
     String flightId = resourceAndFlight.cleanupFlight().flightId();
-    FlightState flightState;
+    Optional<FlightStatus> flightStatus;
     try {
-      flightState = stairway.getFlightState(flightId);
+      flightStatus = Optional.of(stairway.getFlightState(flightId).getFlightStatus());
+    } catch (FlightNotFoundException e) {
```
yonghaoy, line 203: I don't quite understand when and where FlightNotFoundException is throwed. Does line 202 throw this?

wchamber: Yes. Stairway#getFlightState can throw an unchecked FlightNotFoundException if it doesn't have a record of the flightId. That makes sense given the type signature (what else should it return if it doesn't find a flight? arguably null), but I don't think it's great that nothing about that is documented in the function's type signature.

yonghaoy (approved):

PR #73, opened 2020-09-25
Fix swagger generated package name typo

yonghaoy:
We are missing janitor in package name


wchamber (approved):
Did we get it right for the client library?

PR #75, opened 2020-09-28
Add transactions for bump & abandon

wchamber:
- Add transactions for bump and abandon
- Always bump ERROR resources first.


src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
```diff
@@ -121,35 +121,76 @@ public TrackedResourceInfoList getResources(CloudResourceUid cloudResourceUid) {
...
   }
 
   /**
    * Updates the ABANDONED/ERROR resource state to READY.
    *
-   * <p>It is possible that there might be multiple ABANDONED resources, and we only need to update
-   * the one with last expiration time.
+   * <p>Throws {@link NotFoundException} if there were no resources to update.
+   *
+   * <p>There may be one ERROR resource and multiple ABANDONED resources. We should always update
```
yonghaoy, line 166: If there is an ERROR expired 3 days ago. And there is one got abandoned and will expire in 1 day. 
In this case, we will bump the expired resource and it will get cleaned up soon. Would this causing any issue. 
Now I feel this 'bump' has two meanings: 
1. Restart ERROR flight to see if they can succeed. 
2. Resume abandoned flight because they are ready to cleanup now.
I feel it is confused to use the same endpoint for two things. I think you are working on the adding filter to this endpoint, maybe we also need to add the current state as filter, then we know what kind of resource admin want to bump?

wchamber: Thinking more about this, there should never be multiple resources to BUMP. The resources should have been DUPLICATED earlier. The only reason I got my test to pass was that I was accidentally changing the state from DUPLICATED->ERROR. Given that, I think this is ok to use as the single endpoint. Thanks!

yonghaoy: Ah, that make sense. 
One optional(because I think that's too much work, not sure if we need) thing we can do:
Now the logic rely on how we handle duplicate logic. If one day, we change our duplicate logic like not to mark ERROR/ABANDON to duplicate. And we forgot the logic here, it will break this bump endpoint. Do we need to add some test like: (1) insert one resource (2) Mark it as ERROR, (3) insert the same resource again, (4) Abandon it. (5)  bump the resource and it should work fine. 
That is totally optional, we can also just add one line comment in createAndUpdateDuplicate() to reminds us there are  other code assume that for each resource there is only one READY+ERROR+ABANDON. Up to you

yonghaoy (approved):

src/main/java/bio/terra/janitor/service/janitor/TrackedResourceService.java
wchamber: Added a comment. I think it would be difficult to try to write tests to catch against many different future changes. I don't know that ERROR is any different from other states, so I'm not inclined to add a special case for it. We do already have a test for duplicated resources being bumped.

PR #77, opened 2020-09-29
Extend get resource endpoint for resource state filters, limit, and offset

wchamber:
- Extend get resource endpoint
- Refactor TrackedResourceService to deal in internal Java objects instead of the request/response model.


src/main/java/bio/terra/janitor/db/TrackRequest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import bio.terra.janitor.generated.model.CloudResourceUid;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
+import java.time.Instant;
+import java.util.Map;
+
+/** A request to track a resource. This is a precursor to having a {@link TrackedResource}. */
+@AutoValue
+public abstract class TrackRequest {
```
yonghaoy, line 11: Those AutoValue classes under this directory is really confusing to me. I was trying to understand the use of each classes recently when I worked on RbsDao, then lose track. 
Initially we have this: https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/java/bio/terra/janitor/db/TrackedResource.java 
And I think that make sense to have a java class mapped to db schema object in db layer, and better to differentiate with API layer objects. By then I thought there will be only `number of db tables` java object. But then for each db query result we add a new object class, that make sense as well because it is clearer/simpler to return one single object for a `db retrieve method`. But now, there are so many similar objects, and I feel they are just random combination of `CloudResourceUid, State, creation, trackedResourceId`.  I don't think that is easier for people to understand the usage/difference of each of them. Is there anything we can do about this?

src/main/java/bio/terra/janitor/service/janitor/ModelUtils.java
```diff
@@ -0,0 +1,66 @@
...
+import bio.terra.janitor.generated.model.TrackedResourceInfo;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.BiMap;
+import com.google.common.collect.EnumBiMap;
+import com.google.common.collect.ImmutableMap;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+
+/** Helper class for converting to and from the request/response model format. */
+public class ModelUtils {
```
yonghaoy, line 18: One unrelated topic, we have to write a copier function for each API layer - DB layer mapping. Is there library we can use? e.g., in protobuf, there are bunch of proto copier class I have used. 
Found this:, not sure if this is helpful https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/BeanUtils.html#copyProperties-java.lang.Object-java.lang.Object-
I will give it a shot in RBS..

src/main/java/bio/terra/janitor/db/TrackRequest.java
wchamber: We talked some offline about this. The `AndLabels` & `andFlight` TrackedResource classes are a little unfortunate, but there's not another great way to make Pai<TrackedResource, Thing> in Java, and we don't want to have Optional<Thing> that are sometimes not set.
TrackRequest is outside of what's represented in the database as it's a precursor. We don't have a better idea at the moment.

src/main/java/bio/terra/janitor/service/janitor/ModelUtils.java
wchamber: There's AutoConverter at Google. I don't know if there are open source stuff. I generally find that while writing the converters does take 5 minutes of boilerplate, it's not particularly difficult to create or maintain. With some simple test coverage, it's hard to mess up. Whereas magic copier classers still need the simple test coverage and have a little more of a learning curve. But that's all definitely personal opinion and it depends on context.

PR #80, opened 2020-10-06
[No Ticket] Fix few small things while I read Janitor code and increase timeout for deleting project

yonghaoy:


src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleProjectCleanupStep.java
```diff
@@ -57,4 +72,16 @@ protected StepResult cleanUp(CloudResourceUid resourceUid) {
...
+  private static HttpRequestInitializer setHttpTimeout(
+      final HttpRequestInitializer requestInitializer) {
+    return new HttpRequestInitializer() {
+      @Override
+      public void initialize(HttpRequest httpRequest) throws IOException {
+        requestInitializer.initialize(httpRequest);
+        httpRequest.setConnectTimeout(3 * 60000); // 3 minutes connect timeout
+        httpRequest.setReadTimeout(3 * 60000); // 3 minutes read timeout
+      }
+    };
```
wchamber, line 85: I think using Google's HttpRequestInitializer is the right solution. I think this is in the realm of, "Google lets you configure this how you want, and it solves this problem. Use Google's solution: we don't need to wrap or build something for CRL."
For now, if this is the only place we're setting up a Cow that uses HttpRequestInitializer/the only place where we use CloudResourceManagerCow, it's ok. I could see us using dependency injection to coordinate HttpRequestInitializers in the long term, or other services.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleProjectCleanupStep.java
```diff
@@ -57,4 +72,16 @@ protected StepResult cleanUp(CloudResourceUid resourceUid) {
...
   }
+
+  private static HttpRequestInitializer setHttpTimeout(
+      final HttpRequestInitializer requestInitializer) {
+    return new HttpRequestInitializer() {
+      @Override
+      public void initialize(HttpRequest httpRequest) throws IOException {
+        requestInitializer.initialize(httpRequest);
+        httpRequest.setConnectTimeout(3 * 60000); // 3 minutes connect timeout
+        httpRequest.setReadTimeout(3 * 60000); // 3 minutes read timeout
```
wchamber, line 83: Add some comment justification for why we do this.

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleProjectCleanupStep.java
```diff
@@ -24,9 +30,18 @@ public GoogleProjectCleanupStep(ClientConfig clientConfig, JanitorDao janitorDao
   @Override
   protected StepResult cleanUp(CloudResourceUid resourceUid) {
     CloudResourceManagerCow resourceManagerCow;
+
     try {
       resourceManagerCow =
-          CloudResourceManagerCow.create(clientConfig, GoogleCredentials.getApplicationDefault());
+          new CloudResourceManagerCow(
+              clientConfig,
+              new CloudResourceManager.Builder(
```
wchamber, line 38: le'ts also set the application name, like CRL does for the factory function.
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/348bb8896893b8374e37ee1935d6eff07b82a0bb/google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java#L50

wchamber (changes requested):

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleProjectCleanupStep.java
yonghaoy: Yup, I think just do it here is okay for Janitor now. 
I am sure RBS will need DI for this.

wchamber (approved):

PR #83, opened 2020-12-17
[no ticket] Set stairway name and keep log in Janitor

yonghaoy:
Some findings in Buffer service which might also be helpful for Janitor.
Having a fixed name helps us to recover previous stairyway instance since they reuse the same name
Having log helps us debug errors

### DIFF
--- a/src/main/java/bio/terra/janitor/service/stairway/StairwayComponent.java
+++ b/src/main/java/bio/terra/janitor/service/stairway/StairwayComponent.java
@@ -45,6 +45,7 @@ public class StairwayComponent {
         stairwayConfiguration.getClusterName(),
         stairwayConfiguration.getClusterName());
     // TODO(CA-941): Configure the workqueue pubsub subscription and topic for multi-instance.
+    // TODO(PF-314): Cleanup old flightlogs.
     Stairway.Builder builder =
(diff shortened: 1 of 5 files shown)


src/main/resources/application.yml
```diff
@@ -28,6 +28,7 @@ janitor:
     force-clean-start: false
     max-parallel-flights: 40
     migrate-upgrade: true
+    name: janitor-stairway
```
wchamber, line 31: ditto other PR comment about moving this to the default value and explaining its importance.

src/main/java/bio/terra/janitor/service/stairway/StairwayComponent.java
```diff
@@ -49,6 +49,7 @@ public StairwayComponent(
         Stairway.newBuilder()
             .maxParallelFlights(stairwayConfiguration.getMaxParallelFlights())
             .applicationContext(applicationContext)
+            .keepFlightLog(true)
```
wchamber, line 52: ditto other PR TODO for eventual cleanup.

wchamber (approved):
Please also explain why these changes are helpful in the PR description. The most important component of a PR description is the "why," not the "what" of the change.

PR #88, opened 2021-02-08
[PF-438] Bump common-lib and Stairway version

yonghaoy:


wchamber (commented):
gradle lockfile?
what's up with the unit test failure?

yonghaoy:
> gradle lockfile?
> what's up with the unit test failure?

github was down. It should be back now

wchamber (approved):
thanks!

PR #89, opened 2021-02-22
PF-276 - Remove migirate package; rename test databases

ddietterich:
This switches to using the common migrate package.
I used my local installed database system and ran into database name conflicts, so I renamed the test databases.
Running local also exposed PF-485.


wchamber (approved):
LGTM, provided we can figure out the unit test failure oddness.

yonghaoy (approved):

yonghaoy:
> LGTM, provided we can figure out the unit test failure oddness.

It works for me without renaming the db. Two guess:

1. Our FlightschedullerTest eats too many connections
2. `testdb` is a special thing in postgres, and it gets more connections by default.

ddietterich:
Renaming the db causes problems for an unknown reason.

PR #92, opened 2021-02-22
[PF-276] - Test minimum change

ddietterich:
Let's use this PR for PF-276. It contains just the relevant change.

I don't know why a db rename causes unit test failures. I think the unit tests are on thin ice and we'll have to deal with PF-485


yonghaoy (approved):
nit: Fix PR title

PR #93, opened 2021-03-04
[PF-485] Registering connection pool with JMX leads to leaks when @DirtiesContext is used.

jmczerk:
Workaround to prevent connection leak while tests make heavy use of `@DirtiesContext`, see [PF-485] for more detail.

Also includes unrelated fixup of Javadoc errors, and adds javadoc to pre-merge checks (do not squash).

[PF-485]: https://broadworkbench.atlassian.net/browse/PF-485

### DIFF
--- a/src/main/java/bio/terra/janitor/app/configuration/JdbcConfiguration.java
+++ b/src/main/java/bio/terra/janitor/app/configuration/JdbcConfiguration.java
@@ -10,15 +10,37 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 import org.apache.commons.pool2.ObjectPool;
 import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
 /** Base class for accessing JDBC configuration properties. */
 public class JdbcConfiguration {
+  private boolean jmxEnabled = true;
   private String uri;
   private String username;
   private String password;
 
   // Not a property
   private PoolingDataSource<PoolableConnection> dataSource;
+
+  /**
+   * Returns a boolean indicating whether JMX should be enabled when creating connection pools. If
+   * this is enabled, references to connection pools created at context initialization time will be
+   * placed into an MBean scoped for the lifetime of the JVM. This is OK in production, but can be
+   * an issue for tests that use @DirtiesContext and thus repeatedly initialize connection pools, as
+   * these pools will hold active DB connections while held by JMX. Having this configuration allows
+   * us to disable JMX in test environments. See PF-485 for more details.
+   */
+  public boolean isJmxEnabled() {
+    return jmxEnabled;
+  }
+
+  /**
+   * Sets the value for {@link #isJmxEnabled()}. Used by @ConfigurationProperties in extending
+   * classes ({@link JanitorJdbcConfiguration} and {@link StairwayJdbcConfiguration}).
+   */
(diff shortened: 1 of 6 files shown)


yonghaoy (approved):

src/main/java/bio/terra/janitor/app/configuration/JdbcConfiguration.java
```diff
@@ -65,8 +66,11 @@ private void configureDataSource() {
     PoolableConnectionFactory poolableConnectionFactory =
         new PoolableConnectionFactory(connectionFactory, null);
 
+    GenericObjectPoolConfig<PoolableConnection> config = new GenericObjectPoolConfig<>();
```
yonghaoy, line 69: you can add a variable like
private Boolean jmxEnabled = false;

Then JANITOR_DB_JMX_ENABLED becomes an app config that we can easily set.

jmczerk: Good idea; could we then default it `true` and make unit tests configure it `false`, which would then make this a test-only change?

Also, can `JanitorJdbcConfiguration` change this, given it would need to be in the base class (`JdbcConfiguration`) to be available in `configureDataSource()`? 

(`JanitorJdbcConfiguration` is annotated with `@ConfigureProperties`, but `JdbcConfiguration` is not)

yonghaoy: I am not sure if we need Jmx Enabled in 'prod' server or not, and I don't know if we are using that to monitoring things anywhere. So I am not sure what is the default value should be. But default to `true`
 and only `false` in test sounds good.

Sorry, just saw this class is not `@ConfigurationProperties `. So we may need override method someing like `getJmxEnabled()`. 

Actaully, can you try just to put:
`spring.jmx.enabled=false` in `application-test.yml`

This comment is totally optional, I don't think we need jmx, should be fine to just disable it everywhere. 
It is up to you, or see what Will suggests.

jmczerk: Actually, adding the parameter in `JdbcConfiguration` works just fine, since `@ConfigureProperties` just uses the setter defined in the base class.  This allows it to work for `STAIRWAY_DB...` too.

Thanks for this suggestion, I prefer not changing the normal behavior given this is working around something we might remove.

yonghaoy: adding `spring.jmx.enabled=false` in `application-test.yml` will not change normal behaviour. the **application-test** is only used in test.

wchamber: I like Yonghao's suggestion. If we can avoid changing the JdbcConfiguration and control JMX enabled by an existing spring property `spring.jmx.enabled` then not doing a code change for this test sgtm. Still needs your lovely comments in application-test.yml.

jmczerk: I had tried `spring.jmx.enabled=false` before, but that still exhibits the problem.  I suspect that this config parameter is controlling how Spring interacts with JMX (i.e. registering it's own components for management), rather than disabling it.

I agree that making this pure config would be the best solution; I'll look a little more to see if this is an option, but will probably time-box this research.

wchamber (approved):
time-box a non-code alternative sgtm. this is a very reasonable solution, so let's not use too much time on a minor improvement.

yonghaoy (approved):

PR #94, opened 2021-03-10
[PF-284]: Consume new terra-common-lib iam library and exceptions.

jmczerk:
This code removes the local impl of `AuthenticatedUserRequest` and `ProxiedAuthenticatedUserRequestFactory` and uses the common library impl instead.  Additionally removes some associated exceptions which duplicated exceptions in the common library.

From inspection, my read is that none of the Stairway Flight data created by the Janitor persists neither `AuthenticatedUserRequest` nor any of the removed exception types, which would cause backward compatibility issues with active flights upon upgrade.  My read is that the only data types being persisted in Flights by Janitor are `TrackedResourceId` and `CloudResourceUid` ([see `FlightSubmissionFactoryImpl`](https://github.com/DataBiosphere/terra-resource-janitor/blob/11a0b0a7bc0ef441023dc9253be60a43fc71e88b/src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java#L18)).  I would appreciate if reviewers have insight to confirm or deny this.


wchamber (approved):
Yes, Janitor does not serialize the AuthenticatedUserRequest, so there is no backwards compatibility issues. Janitor does not execute any flights for the user.

yonghaoy (approved):

PR #96, opened 2021-03-11
PF-521 Migrate Janitor to use common tracing package

wchamber:
Also make it a log error instaed of an exception if unable to export metrics to stackdriver.


src/main/java/bio/terra/janitor/service/stackdriver/StackdriverExporter.java
```diff
@@ -26,24 +21,14 @@ public StackdriverExporter(StackdriverConfiguration stackdriverConfiguration) {
...
+    logger.info("Stackdriver stats enabled: {}.", stackdriverConfiguration.isEnabled());
     if (!stackdriverConfiguration.isEnabled()) {
       return;
     }
     try {
       StackdriverStatsExporter.createAndRegister();
-      StackdriverTraceExporter.createAndRegister(StackdriverTraceConfiguration.builder().build());
     } catch (IOException e) {
-      throw new RuntimeException("Unable to initialize Stackdriver exporting.", e);
+      throw new RuntimeException("Unable to initialize Stackdriver stats exporting.", e);
```
yonghaoy, line 31: do you also switch to `logger.error` instead of exception?

yonghaoy (approved):

src/main/java/bio/terra/janitor/service/stackdriver/StackdriverExporter.java
wchamber: thanks, done.

jmczerk (approved):

PR #106, opened 2021-05-20
PF-194  Add resource metadata support.

wchamber:
Add resource metadata to the client, internal representation, and database to allow adding additional information about a tracked resource that isn't its unique id or a label.
We will initially use this for Google project parents in order to help cleanup of projects that don't exist.


yonghaoy (approved):
LG. You may also want to bump version in build.gradle before publish client library

PR #107, opened 2021-05-20
PF-194 Check folder permissions during project cleanup

wchamber:
If Janitor has permission on that folder but get 403 when getting project, that means project doesn’t exists. Allow the cleanup to complete successfully in that case. Use the googleParentResource metadata for this, if it exists.


src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleUtils.java
```diff
@@ -2,15 +2,95 @@
...
+
+  /**
+   * Our best determination of a project that may be deleted status. See {@link
+   * #checkProjectStatus(String, Optional, CloudResourceManagerCow)}.
+   */
+  enum ProjectStatus {
+    // The project is exists and is active.
+    ACTIVE,
+    // The project is in the process of being deleted. It is partially unavailable.
+    DELETE_IN_PROGRESS,
```
yonghaoy, line 86: GCP Lifecycle also has this name. And we use the same name but mapped to two GCP status:
`DELETE_REQUESTED` and `DELETE_IN_PROGRESS`. This confuses me if I did not trace whether its Janitor or GCP's  `DELETE_IN_PROGRESS`. Using different name to summarize those two GCP status might be helpful?  Like `DELETING`?

yonghaoy (approved):

src/main/java/bio/terra/janitor/service/cleanup/flight/GoogleUtils.java
wchamber: Good idea, done.

PR #112, opened 2021-06-02
[PF-762] Consume CRL Gradle Platform.

jmczerk:
Consume the [Gradle Platform](https://docs.gradle.org/current/userguide/java_platform_plugin.html) from CRL.  Note this involves an implicit bump of CRL `google-storage` library from `0.11.0` to `0.12.0`.


yonghaoy (approved):

build.gradle
```diff
@@ -59,11 +59,13 @@ dependencies {
     // Terra deps
     implementation group: 'bio.terra', name: 'stairway', version: '0.0.52-SNAPSHOT'
     implementation group: "bio.terra", name: "terra-common-lib", version: "0.0.46-SNAPSHOT"
+
+    implementation platform('bio.terra.cloud-resource-lib:platform:0.1.0')
     implementation group: 'bio.terra.cloud-resource-lib', name: 'cloud-resource-schema'
```
wchamber, line 64: we define a `cloud-resource-schema` version above to  pull the openapi definitions for generating the janitor client. While we probably want to eventually get rid of that, for now should we no longer define the version on line 35 and get it from the platform?

wchamber (approved):

build.gradle
jmczerk: Ah yes, good catch, thanks.  I should have noticed that it didn't have a local version specified.

PR #118, opened 2021-10-05
[PF-980] Upgrade to new Stairway and Common Lib

ddietterich:
Changes in this PR:
- Update to new TCL
- Remove stairway dependency; resolve it via TCL
- Stairway no longer supports flight delete. Remove it from fatal flight handling. Also, use mockito spy in place of flight delete for handling lost flights
- Enable background flight cleaning. Currently set to check daily and retain flights for one week.


src/main/resources/application.yml
```diff
@@ -30,6 +30,8 @@ janitor:
     migrate-upgrade: true
     quiet-down-timeout: 20s
     terminate-timeout: 5s
+    retention-check-interval: 1d
+    completed-flight-retention: 7d
```
yonghaoy, line 34: can we make this little longer? It took 30 days for Google to remove delete a project. It might be helpful if we can keep the light little longer than that.

yonghaoy (approved):

zloery (approved):

PR #119, opened 2021-11-15
[no ticket] add readme to explain how to connect to tools Janitor DB

yonghaoy:


wchamber (approved):
It'd be nice if it weren't necessary to connect directly to the DB and we had sufficient dev API coverage. Anything in particular prompting this addition?

yonghaoy:
> It'd be nice if it weren't necessary to connect directly to the DB and we had sufficient dev API coverage. Anything in particular prompting this addition?

There was an ask about why CWB fiab created project is deleted. So I pulled some data from janitor. Checking via DB is hard to use(especially writing sql queries to deal with jsonb format...)
I end up with using swagger UI.

PR #120, opened 2021-11-19
[IA-2938] Clean up Azure IPs

rtitle:
First pass at adding Janitor support for Azure resources. Currently just cleans up `PublicIp`, but if we find this approach works we can start adding other resources.

As part of this I also updated Janitor to pull Azure creds from vault. It currently works for local/integration testing; we'll probably need to update `terra-helm` to pull the new vault secrets for this to go live.


build.gradle
```diff
@@ -83,6 +85,9 @@ dependencies {
     implementation group: 'javax.validation', name: 'validation-api', version: '2.0.1.Final'
     implementation group: 'org.liquibase' , name: 'liquibase-core', version: '3.10.0'
     runtimeOnly group: 'org.postgresql', name: 'postgresql', version: '42.2.12'
+    // TODO: consider changing these dependencies to `api` in CRL so they get pulled in transitively
```
rtitle, line 88: Hopefully resolved with https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/101

build.gradle
```diff
@@ -8,7 +8,7 @@ plugins {
     id 'de.undercouch.download' version '4.0.0'
     id 'io.spring.dependency-management' version '1.0.10.RELEASE'
     id 'org.hidetake.swagger.generator' version '2.18.2'
-    id 'org.springframework.boot' version '2.3.1.RELEASE'
```
rtitle, line 11: 2.3.1 had some issues with the Azure libs -- 2.4.0 works fine, and is what WSM uses.

local-dev/render-config.sh
```diff
@@ -9,10 +9,13 @@ VAULT_SERVICE_ACCOUNT_PATH=secret/dsde/terra/kernel/integration/toolsalpha/crl_j
 VAULT_CLIENT_SERVICE_ACCOUNT_PATH=secret/dsde/terra/kernel/integration/toolsalpha/crl_janitor/client-sa
 VAULT_TOOLS_CLIENT_SERVICE_ACCOUNT_PATH=secret/dsde/terra/kernel/integration/tools/crl_janitor/client-sa
 VAULT_CLOUD_ACCESS_SERVICE_ACCOUNT_PATH=secret/dsde/terra/janitor-test/default/cloud-access-sa
+VAULT_AZURE_MANAGED_APP_CLIENT_PATH=secret/dsde/terra/kernel/integration/tools/crl_janitor/azure-managed-app-client
```
yonghaoy, line 12: how did you pull those secrets into vault? 
Can we put that into terraform:
https://github.com/broadinstitute/terraform-ap-modules/blob/master/crl-janitor/outputs.tf
https://github.com/broadinstitute/terraform-ap-deployments/blob/master/terra-env/vault_crl_janitor.tf

src/main/java/bio/terra/janitor/app/configuration/CrlConfiguration.java
```diff
@@ -69,4 +83,28 @@ private static HttpRequestInitializer setHttpResourceManagerTimeout(
       httpRequest.setReadTimeout(3 * 60000); // 3 minutes read timeout
     };
   }
+
+  /** Creates an Azure {@link ComputeManager} client for a given managed resource group. */
+  public ComputeManager buildComputeManager(AzureResourceGroup azureResourceGroup) {
```
yonghaoy, line 88: We might be able to create a `ComputeManager` Bean just like the `Cows` above?  
Then subclasses can inject that bean directly.

yonghaoy (commented):
Change looks good, haven't review the CRL change yet. 
One thing to note, the "production janitor" than runs in GKE uses this chart:
https://github.com/broadinstitute/terra-helm/blob/master/charts/crljanitor/templates/deployment.yaml#L64L105
(the `render_config` is just for local testing)

local-dev/render-config.sh
yonghaoy: Also in the same Terraform module, we grant Janitor DELETE permissions on those GCP folders. Can we also have that for Azure.

yonghaoy (approved):

src/main/java/bio/terra/janitor/app/configuration/AzureConfiguration.java
```diff
@@ -0,0 +1,39 @@
+package bio.terra.janitor.app.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "janitor.azure")
+public class AzureConfiguration {
```
ddietterich, line 10: I'm unsure about this structure. This would have a Janitor configured to clean up a specific managed app installation in a tenant.

What will we do when there is more than one installation in an environment?

ddietterich (commented):
I'm not familiar with the Janitor code, but gave this a look.
I have a question.

src/main/java/bio/terra/janitor/app/configuration/CrlConfiguration.java
rtitle: I thought about that but we need a `ComputeManager` instance for each subscription/resource group. Since Janitor will need to clean up resources across subscriptions, I'm building it on-demand instead of making it a singleton `Bean`.

local-dev/render-config.sh
rtitle: > Can we put that into terraform
Yes! Will make a PR to terraform this. (I wrote a single secret manually but that's probably not best practice :)).

> we grant Janitor DELETE permissions on those GCP folders. Can we also have that for Azure.
Hm, I think the Azure model is slightly different. Janitor will have the identity of a managed app _publisher_, which can delete resources in a managed app, but can't delete the managed app itself. I'm picturing we'll have a static managed app deployment used for integration testing.

src/main/java/bio/terra/janitor/app/configuration/AzureConfiguration.java
rtitle: Technically Janitor here has the identity of a managed app _publisher_. So it can clean up deployments of that managed app across tenants/subscriptions. The tenant/subscription/mrg is included in the `CloudResourceUid` message.

Even if we're working with multiple managed app _definitions_ it should work I think -- we would just need to add the Janitor `clientId` in the Terra tenant to the managed app publisher group.

ddietterich: Got it, thanks

ddietterich (approved):
LGTM

rtitle:
Updated this with the latest CRL and tested end to end. I think this is good to go but I'll hold off from merging until next week -- I still need to create the `terra-helm` PR for the 'live' deploy..

rtitle:
Couple updates here:
1. I synced with DevOps and they are looking at adding Terraform support for Azure envs (see https://broadworkbench.atlassian.net/browse/DDO-1772).
2. In the meantime I manually inserted a vault secret for the managed app publisher and updated the `crljanitor` chart in terra-helm to use it: https://github.com/broadinstitute/terra-helm/pull/645

So I think once the above `terra-helm` change is in, then this PR can be merged. I will then make a separate Janitor PR to support other Azure resources (disk, network, vm).

PR #121, opened 2021-11-24
[no ticket] bump cloud resource uid version to have Azure changes

yonghaoy:
https://broadworkbench.atlassian.net/browse/PF-866, for cloud resource uid, we need to manually upgrade the version, 
Currently Janitor is broken due to invalid messages.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ allprojects {
             //
             // https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/211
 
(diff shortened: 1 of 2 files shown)


rtitle (approved):
Looks good, thanks.

What do you mean Janitor is currently broken? `cloud-resource-schema:0.10.0` has the new Azure resources, but nothing is publishing them yet, until https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/101/ is used. So I had assumed Janitor would still function -- just wondering if I broke anything.

PR #122, opened 2021-12-06
IA-2938 Clean up Azure network, disk, VM 

rtitle:
This follows up on https://github.com/DataBiosphere/terra-resource-janitor/pull/120 and adds Azure network, disk, VM resources to Janitor cleanup.

Integration tests pass; will test end-to-end with WSM as well.


src/main/java/bio/terra/janitor/service/cleanup/flight/AzureNetworkSecurityGroupCleanupStep.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  protected StepResult cleanUp(CloudResourceUid resourceUid, ResourceMetadata metadata)
+      throws InterruptedException, RetryException {
+    AzureNetworkSecurityGroup networkSg = resourceUid.getAzureNetworkSecurityGroup();
+    ComputeManager computeManager =
+        crlConfiguration.buildComputeManager(networkSg.getResourceGroup());
+
+    try {
```
yonghaoy, line 36: optional: We might be able to exact a common method that execute a method and return a step? 
Example:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/resource/flight/GoogleUtils.java#L100-L109

yonghaoy (approved):

src/main/java/bio/terra/janitor/service/cleanup/flight/AzureNetworkSecurityGroupCleanupStep.java
rtitle: Yeah.. I tried to do something similar at first but there are some slight differences and it was a bit of a pain. Maybe will revisit though since there is some duplicated code.

Qi77Qi (approved):
lgtm...is next step we'll update WSM to use this?

rtitle:
> is next step we'll update WSM to use this?

Yes, we should update WSM to use the `terra-integration-test` MRG and read secrets from vault. Also update update CI to run the `azure-test` tests (or maybe remove the `azure-test` spring profile entirely?). I kinda want `azure-poc` to be rebased before we do that though..

src/main/java/bio/terra/janitor/service/cleanup/flight/AzureNetworkSecurityGroupCleanupStep.java
yonghaoy: We might be able to do something similar to the example above.
1: Define a FunctionalInterface 
```
  @FunctionalInterface
  public interface StairwayExecute {
    void execute();
  }
```
Then a util method: 
```
  public static StepResult DeleteAzureResourceAndIgnoreNotFound(StairwayExecute execute) {
    try {
      execute.execute();
    } catch (ManagementException e) {
            if (StringUtils.equals(e.getValue().getCode(), "ResourceNotFound")) {
        logger.info(
            "Azure network security group {} in managed resource group {} already deleted",
            networkSg.getNetworkSecurityGroupName(),
            networkSg.getResourceGroup().getResourceGroupName());
        return StepResult.getStepResultSuccess();
    }
  }
```

yonghaoy (approved):

src/main/java/bio/terra/janitor/service/cleanup/flight/AzureNetworkSecurityGroupCleanupStep.java
rtitle: Pushed up a change to do this :)

PR #124, opened 2021-12-14
[PF-1251] Bump log4j api to 2.16.0

zloery:
Bump log4j-api transitive dependency version from 2.15.0 to 2.16.0. Also switches to the Spring dependency management plugin syntax.


ungwudik (approved):

build.gradle
```diff
@@ -106,18 +107,6 @@ dependencies {
 
     annotationProcessor group: 'com.google.auto.value', name: 'auto-value', version: '1.7.3'
     annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor'
-
-    // Force log4j version to at least 2.15.0 for security concerns.
-    implementation('org.apache.logging.log4j:log4j-api') {
```
yuhuyoyo, line 111: why are we not changing it to 2.16 here like what you did in the other repo?

build.gradle
```diff
@@ -16,6 +16,7 @@ allprojects {
     group = 'bio.terra'
     version = System.getenv('SERVICE_VERSION') != null ? System.getenv('SERVICE_VERSION') : '0.5.0-SNAPSHOT'
     sourceCompatibility = JavaVersion.VERSION_11
+    ext['log4j2.version'] = '2.16.0' // until Spring Boot 2.6.2
```
yuhuyoyo, line 19: what does until spring boot 2.6.2 means?

yuhuyoyo (commented):
Thanks for fixing this. I left some questions

build.gradle
zloery: This does bump us to 2.16.0, but this adds the constraint via the Spring dependency manager plugin (https://docs.spring.io/spring-boot/docs/2.6.1/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.customizing) instead of directly. 
Mostly it's just less verbose, the end effect is the same here

build.gradle
zloery: We currently pull in the log4j-api jars as transitive dependencies of Spring (and possibly other dependencies as well). Spring has said they'll upgrade their dependency version in the 2.6.2 release (expected late December 2021, https://spring.io/blog/2021/12/10/log4j2-vulnerability-and-spring-boot), so at that point we may not need to explicitly override their version here.

PR #131, opened 2022-03-08
[PF-1228] Fixup dependency removals and consume latest CRL.

jmczerk:
The [original PR](https://github.com/DataBiosphere/terra-resource-janitor/pull/126/files) to implement PF-1228 missed a couple items that needed to be removed, specifically around usage of the no-longer-existing `cloud-resource-schema` library.  

This was masked by the fact that we were using an old, mismatched version of CRL.  This PR removes the bad references and upgrades to latest CRL version.  Note that this aligns more closely with the [RB PR for PF-1228](https://github.com/DataBiosphere/terra-resource-buffer/pull/216).


build.gradle
```diff
@@ -91,6 +89,7 @@ dependencies {
     swaggerCodegen group: 'io.swagger.codegen.v3', name: 'swagger-codegen-cli'
     runtimeOnly group: 'org.webjars.npm', name: 'swagger-ui-dist', version: '3.35.2'
     implementation group: 'io.swagger.core.v3', name: 'swagger-annotations'
+    implementation group: 'io.swagger', name: 'swagger-annotations', version: '1.6.2'
```
zloery, line 92: I think this version is managed by the Spring dependency plugin block above, so you shouldn't need to specify the version here or in `terra-resource-janitor-client/build.gradle`

zloery (approved):

build.gradle
jmczerk: Good catch, thanks!

sergiygetlin (approved):

PR #136, opened 2022-04-14
[PF-1557] Follow WSM pattern for publishing

zloery:
Janitor's `tag-build-push` workflow is currently broken because it uses `broadinstitute/github-tag-action` instead of `databiosphere/github-actions/actions/bumper`. The old action hasn't been updated since 2020, and recently broke due to git's mitigation of CVE-2022-24765 (see example workaround in https://github.com/DataBiosphere/github-actions/pull/32). Rather than fix a second bumper action, I think it makes sense to use the action currently supported by Broad devops.

This change includes:

- Swapping `broadinstitute/github-tag-action` to `databiosphere/github-actions/actions/bumper`, plus a handful of related changes inside the `tag-build-push` workflow inspired by [WSM's workflow](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/.github/workflows/tag-publish.yml).
- Adding a top-level gradle version property to `settings.gradle`
- Adding a local `bump-skip` action copied from WSM. We should probably move this to a central place instead of copying it everywhere like we currently do, but I'm not sure the right way to do that.

### DIFF
--- a/.github/workflows/master_push.yml
+++ b/.github/workflows/master_push.yml
@@ -122,6 +122,7 @@ jobs:
         if: steps.skiptest.outputs.is-bump == 'no'
         run: "./gradlew jib --image=gcr.io/${GOOGLE_PROJECT}/${SERVICE_NAME}:${{ steps.tag.outputs.tag }}"
       - name: Update Version Mapping
+        if: steps.skiptest.outputs.is-bump == 'no'
         uses: broadinstitute/repository-dispatch@master
         with:
           token: ${{ secrets.REPO_ACCESS_TOKEN }}
--- a/.github/workflows/master_push.yml
+++ b/.github/workflows/master_push.yml
@@ -62,7 +62,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOTFIX_BRANCHES: hotfix.*
           OVERRIDE_BUMP: ${{ steps.controls.outputs.semver-part }}
-          RELEASE_BRANCHES: msdtrt
+          RELEASE_BRANCHES: master
           VERSION_FILE_PATH: settings.gradle
           VERSION_LINE_MATCH: "^gradle.ext.janitorVersion\\s*=\\s*\".*\""
           VERSION_SUFFIX: SNAPSHOT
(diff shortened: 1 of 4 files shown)


jaycarlton:
Maybe you could check it into the databiosphere actions repo.

.github/workflows/master_push.yml
```diff
@@ -17,18 +31,47 @@ jobs:
...
-      - name: Bump version and push tag
+        with:
+          ref: ${{ steps.controls.outputs.checkout-branch }}
+      - name: Skip version bump merges
+        id: skiptest
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
+      - name: Bump the tag to a new version
+        if: steps.skiptest.outputs.is-bump == 'no'
```
jaycarlton, line 57: I wish there were a way around all these if's. Maybe we could split the job into two and only call the second part if it's not a bump. I.e. do the bifurcation on the input filter of the job instead of at all the steps.

.github/workflows/master_push.yml
```diff
@@ -17,18 +31,47 @@ jobs:
...
+          OVERRIDE_BUMP: ${{ steps.controls.outputs.semver-part }}
+          RELEASE_BRANCHES: master
+          VERSION_FILE_PATH: settings.gradle
+          VERSION_LINE_MATCH: "^gradle.ext.janitorVersion\\s*=\\s*\".*\""
+          VERSION_SUFFIX: SNAPSHOT
       - name: Pull Vault image
+        if: steps.skiptest.outputs.is-bump == 'no'
         run: docker pull vault:1.1.0
       # Currently, there's no way to add capabilities to Docker actions on Git, and Vault needs IPC_LOCK to run.
       - name: Get Vault token
```
jaycarlton, line 73: It would be more in line with current best practices to use github secrets.

jaycarlton (approved):

.github/workflows/master_push.yml
ddietterich: If you figure out a way that works, let me know. The continue to add features to GHA, so maybe we can do that someday.

ddietterich (approved):
LGTM

.github/workflows/master_push.yml
zloery: Agreed, though I think it's worth doing that migration separately

PR #138, opened 2022-04-18
[PF-1557] Checkout repo as Broadbot in build-tag-publish workflow

zloery:
Per documentation at https://github.com/actions/checkout, the `checkout` action persists a git token and it seems like this overrides the token provided to the `bumper` action, causing the bumptagbot commit to fail (see "Bump the tag to a new version" [section of GHA run](https://github.com/DataBiosphere/terra-resource-janitor/runs/6067290741?check_suite_focus=true)). 

This also stops tests from running on version bump commits, as they'll always be the same as the real commit which caused them.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,10 +35,17 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Skip version bump merges
+        id: skiptest
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
       - name: Pull Vault image
+        if: steps.skiptest.outputs.is-bump == 'no'
         run: docker pull vault:1.1.0
       # Currently, there's no way to add capabilities to Docker actions on Git, and Vault needs IPC_LOCK to run.
       - name: Get Vault token
+        if: steps.skiptest.outputs.is-bump == 'no'
         id: vault-token-step
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -70,17 +82,21 @@ jobs:
...
+        if: steps.skiptest.outputs.is-bump == 'no'
         id: unit-test
         run: ./gradlew unitTest --scan
       - name: Run integration tests
+        if: steps.skiptest.outputs.is-bump == 'no'
         id: integration-test
         run: ./gradlew integrationTest --scan
       - name: Upload Test Reports
-        if: always()
+        if: always() && steps.skiptest.outputs.is-bump == 'no'
```
jaycarlton, line 99: this hurts my head. Do you still need the `always()`?

jaycarlton (approved):

.github/workflows/test.yml
zloery: Yeah - from some poking around it looks like unless you call `always()` or one of a few similar functions (`success()`, `cancelled()`), [there's an implicit check of `success()`](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions). Calling one of those functions returns a value, but also has the side-effect of overriding this default.

They also hint at that in the [`always()` documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#always) but it would be nice if they explicitly called out that `if: always() && X` behaves differently than `if: X`

PR #139, opened 2022-05-06
[PF-1321] Add support for cleaning up WSM workspaces

zloery:
This change adds a new resource type to Janitor: `TerraWorkspace`. It's uniquely specified by the workspace ID (UUID) and instance identifier (String matching one of Janitor's configuration values). It also has a metadata value field tracking which user owns the workspace.

The actual cleanup is done by impersonating a test user and calling WSM's `deleteWorkspace` endpoint as the user. The `tools` Janitor SA has permission to impersonate test users via domain-wide delegation.

This also involves a few related changes:

- Moving `swagger-ui.html` and `service_openapi.yaml` from `src/resources/api` to `src/resources/static` so they can be accessed directly. This was previously possible, but I think we lost the ability to do this with a Spring upgrade.
- Upgrading ResourceMetadata to v2 by adding a new field. This is backwards compatible with existing serialized V1 instances, which I've verified in unit tests.
- Externally to this PR, I've enabled domain-wide delegation of test users for the Janitor SA and confirmed it can impersonate them properly via manual testing.


src/main/resources/static/service_openapi.yaml
```diff
@@ -288,7 +288,7 @@ components:
         - DONE
         - ERROR
         - ABANDONED
-        - DUPLIATED
+        - DUPLICATED
```
jaycarlton, line 291: Sure fixing this won't break anything?

src/main/resources/static/service_openapi.yaml
```diff
@@ -529,6 +535,19 @@ components:
...
+      - workspaceId
+      - workspaceManagerInstance
+      properties:
+        workspaceId:
+          type: string
+          format: uuid
+          description: The UUID of the workspace
+        workspaceManagerInstance:
+          type: string
+          description: The Workspace Manager instance which manages this workspace
```
jaycarlton, line 550: document that this is the URL not the map key from the other yaml file.

jaycarlton (approved):

melissachang (approved):
Nice!

src/main/java/bio/terra/janitor/service/workspace/WorkspaceManagerService.java
```diff
@@ -0,0 +1,57 @@
...
+  public WorkspaceManagerService(
+      WorkspaceManagerConfiguration workspaceManagerConfiguration, IamService iamService) {
+    this.workspaceManagerConfiguration = workspaceManagerConfiguration;
+    this.iamService = iamService;
+    this.commonHttpClient = new ApiClient().getHttpClient();
+  }
+
+  private ApiClient getApiClient(String accessToken, String basePath) {
+    ApiClient client = new ApiClient().setHttpClient(commonHttpClient);
+    client.setBasePath(basePath).setAccessToken(accessToken);
```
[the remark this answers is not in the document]
zloery: Nope - for some reason `setHttpClient` and `setBasePath` are fluent, but `setAccessToken` isn't. 🤷

src/main/resources/static/service_openapi.yaml
zloery: Thinking about this more, Janitor is returning this to clients, so any old client version talking to the new server would probably have trouble parsing the enum. 
We could either make a breaking change or do a two part "introduce new enum, then delete the old one" migration. Either way, probably better to do that on it's own, PF-1628 for when someone needs a bite-sized change

src/main/java/bio/terra/janitor/db/JanitorDao.java
```diff
@@ -439,29 +439,36 @@ static ResourceMetadata deserializeMetadata(@Nullable String resource) {
...
 
-  /** POJO class to use for JSON serializing a {@link ResourceMetadata}. */
+  /**
+   * POJO class to use for JSON serializing a {@link ResourceMetadata}. This adds the workspaceOwner
+   * field to the MetadataModelV1 class, making it backwards (but not forwards) compatible with
+   * version 1.
+   */
   @VisibleForTesting
-  static class MetadataModelV1 {
+  static class MetadataModelV2 {
```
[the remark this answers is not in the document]
zloery: SGTM. I'm still going to bump the serialized "version" field and leave the comment, but good point about not needing to write a new class.

PR #140, opened 2022-05-10
[PF-883] Store WSM url allowlist in Janitor config instead of using identifiers

zloery:
Mapping identifiers to WSM urls meant we needed to keep hardcoded strings in sync across services, using urls in requests avoids this. However, we're still using a config-based allowlist, which prevents users from forcing Janitor to call arbitrary urls.

See usage in https://github.com/DataBiosphere/terra-cli/pull/246


src/main/java/bio/terra/janitor/app/configuration/WorkspaceManagerConfiguration.java
```diff
@@ -12,13 +12,13 @@ public class WorkspaceManagerConfiguration {
 
   // A map from instance IDs to Workspace Manager instance URLs, like:
```
melissachang, line 13: Update comment

melissachang (approved):
Thanks

PR #141, opened 2022-06-14
[TOAZ-162] Add Janitor support for managed identities

rtitle:
Friends with: https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/132

Add Janitor support to clean up managed identity resources created via CRL in tests.


src/main/java/bio/terra/janitor/app/configuration/CrlConfiguration.java
```diff
@@ -87,71 +88,46 @@ private static HttpRequestInitializer setHttpResourceManagerTimeout(
   }
 
   /** Creates an Azure {@link ComputeManager} client for a given managed resource group. */
-  public ComputeManager buildComputeManager(AzureResourceGroup azureResourceGroup) {
-    TokenCredential azureCreds =
```
rtitle, line 91: Factored out some duplicated code in this class.

src/test/java/bio/terra/janitor/integration/TrackResourceIntegrationTest.java
```diff
@@ -929,6 +934,39 @@ public void subscribeAndCleanupResource_alreadyDeletedTerraWorkspace() throws Ex
     publishAndVerify(resource, ResourceState.DONE, metadata);
   }
 
+  @Test
+  public void subscribeAndCleanupResource_azureManagedIdentity() throws Exception {
```
rtitle, line 938: Ran locally, this test passed

src/main/resources/static/service_openapi.yaml
```diff
@@ -550,6 +550,17 @@ components:
           description: |
             The URL of the Workspace Manager instance which manages this workspace. This must be
             in Janitor's configured allowlist.
+    AzureManagedIdentity:
```
Qi77Qi, line 553: do we want to specify `userManaged` identity? just to differentiate it from system assigned ones...or at this level, it could be both?

zykonda: The System-assigned managed identities are deleted with resources they've been created on. So, there is no need to handle system-assigned MI's here. As for the naming, I think AzureManagedIdentity is good enough.

Qi77Qi (approved):

src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
```diff
@@ -48,6 +48,8 @@ public FlightSubmission createSubmission(TrackedResource trackedResource) {
         return FlightSubmission.create(AzureContainerInstanceCleanupFlight.class, flightMap);
       case TERRA_WORKSPACE:
```
zykonda, line 49: Once we start to assign managed identities, we need to update this code to remove assignments of MIs to workspace resources as assignments are not deleted automatically upon MI deletion and will be dangling as Unknown references. We need a task or mention for this clean up job.

zykonda (approved):
Looks good

src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java
rtitle: Hm true, MI deletion in Janitor _could_ also find+remove all assignments. However VMs should also be cleaned up by Janitor. Maybe once we have some connected tests in WSM that assign VMs to MIs we can observe how well Janitor works.

PR #142, opened 2022-08-05
Fix GHA secret masking

zloery:
(Same situation as https://github.com/DataBiosphere/terra-resource-buffer/pull/244)
These workflows are masking secrets incorrectly, the masking needs to happen before setting them as output. If this workflow is run with debug logging (which it isn't), the current pattern will leak secrets.


mflinn-broad (approved):

choover-broad (approved):

zloery:
It looks like there might be a broken/flaky test but I'm confident it's unrelated to this change, merging anyways

PR #143, opened 2022-08-12
[PF-1731] Replace Java setup action

zloery:
The `joschi/setup-jdk@v2` Github action is deprecated and the recommendation is to move to the official github setup-java action with the `temurin` distribution: see https://github.com/joschi/setup-jdk#-notice


zloery:
Looks like the `subscribeAndCleanupResource_azureRelayAndHybridConnections` test is broken, #144 should fix it

dexamundsen:
handled along with JDK upgrade in https://github.com/DataBiosphere/terra-resource-janitor/pull/145

PR #156, opened 2023-02-24
[PF-963] Update wsm client version, revert jersey-inject version to match

zloery:
Janitor is currently using version 3.0.8 of the `org.glassfish.jersey.inject:jersey-hk2` library, but version 2.35 of other Jersey libraries. Something about this combination of libraries is incompatible and all Janitor calls to WSM since this was added have silently died. As a result, there are currently 37,318 workspace cleanup flights "in progress" on `tools` Janitor 🙃.

Stairway is configured to run max 40 parallel flights, so this shouldn't overwhelm the dev WSM server too badly. Additionally, most of these workspaces have likely been deleted already, so many responses should be a quick 404.

While debugging this I also noticed the WSM client was very old, so I bumped that as well. I don't believe it's related to this issue, but might as well.


dexamundsen (approved):
I vaguely remember compatibility issues between 3.x and 2.3x versions of jersey. this related to 3.x expecting of jakarta.* packages, but 2.3 still using javax.*

build.gradle
```diff
@@ -85,10 +85,10 @@ dependencies {
     // Terra libraries
     implementation group: 'bio.terra', name: 'terra-common-lib', version: '0.0.76-SNAPSHOT'
     implementation group: 'bio.terra', name: 'terra-cloud-resource-lib', version: '1.2.4-SNAPSHOT'
-    implementation group: 'bio.terra', name: 'workspace-manager-client', version: '0.254.345-SNAPSHOT'
+    implementation group: 'bio.terra', name: 'workspace-manager-client', version: '0.254.575-SNAPSHOT'
 
     // hk2 is required to use WSM client, but not correctly exposed by the client
-    implementation group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '3.0.8'
+    implementation group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '2.35'
```
yuhuyoyo, line 91: shouldn't it be 2.32 if you are matching?
https://github.com/DataBiosphere/terra-workspace-manager/blob/433c236206ded0e101708c5adfb96f34ad0cd8bf/client/gradle/dependencies.gradle#L3

yuhuyoyo: ah you are matching the 2.35 below.

can we define a jerseyVersion variable and share it to prevent future discrepency?

yuhuyoyo (approved):

PR #164, opened 2023-05-26
[PF-2781] Use signed JWT to impersonate test users

zloery:
Currently, Janitor tries to use the application-default credentials (ADC) from running in GKE to impersonate test users (via domain-wide delegation). This works when the ADC come from an exported service account key, but does not work for Workload Identity, as the provided credentials do not support delegation.

Instead, this uses a new approach for Janitor:
1. Pull ADC from the environment
2. Create a custom JWT to impersonate a test user. Use the ADC from the environment to sign this JWT
3. Exchange the signed JWT for an access token authenticated as the test user
4. Use the test user access token normally in Terra


src/main/java/bio/terra/janitor/service/iam/IamService.java
```diff
@@ -60,21 +104,96 @@ public void requireTestUser(String userEmail) {
...
+   * support DWD. Instead, this method builds a custom JWT with the subject claim, uses the ADC to
+   * sign the JWT via the IamCredentials API, and finally exchanges the signed JWT for an OAuth
+   * access token to call WSM.
+   *
+   * @param testUserEmail Email of the test user to impersonate. This will be validated against the
+   *     domain allowed in configuration properties
+   * @return OAuth access token authorizing the caller to impersonate the provided test user.
    */
-  public AccessToken impersonateTestUser(String testUserEmail) {
+  public String impersonateTestUser(String testUserEmail) {
```
dexamundsen, line 119: nit: could rename function to indicate that access token is returned

dexamundsen (approved):

jmczerk (approved):
This is very cool, I didn't realize we could get Google to sign JWT's like this!

PR #168, opened 2023-11-03
TOAZ-336 phase 1: Add missing Azure resources to Janitor

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-336

This PR adds missing Azure resource types to Janitor, and removes some legacy ones.

New resources:
* AzureDatabase
* AzureKubernetesNamespace

Removed resources:
* AzurePublicIp
* AzureNetwork
* AzureNetworkSecurityGroup
* AzureRelay
* AzureContainerInstance

The removed ones are never passed to Janitor today by any service, and are not intended to be since they are not WSM resource types. The code changes here are backwards compatible with existing JSON messages. So the removal is safe, I think.

Note this PR does NOT yet add handling for `AzureDatabase` and `AzureKubernetesNamespace`. That will come in a follow-up PR, after CRL is updated (there is kind of a circular dependency between Janitor and CRL).


src/test/resources/application-integration.yml
```diff
@@ -17,6 +17,12 @@ janitor:
     resource-credential-file-path: rendered/cloud-access-sa-account.json
     resource-project-id: terra-janitor-test
     track-resource-topic-id: crljanitor-toolsalpha-pubsub-topic
-    azure-tenant-id: 0cb7a640-45a2-4ed6-be9f-63519f86e04b
-    azure-subscription-id: 3efc5bdf-be0e-44e7-b1d7-c08931e3c16c
-    azure-managed-resource-group-name: mrg-terra-integration-test-20211118
+    # Reusing static MRG from https://github.com/broadinstitute/terra-github-workflows/blob/main/.github/workflows/attach-billing-project-to-landing-zone.yaml
```
yonghaoy, line 20: fyi (and you probably know already). Those are configs for janitor integration test. 
The "production Janitor" config is in vault: https://github.com/broadinstitute/terra-helmfile/blob/04733580a5fbde7141a4f0378c9763525009c1d1/charts/crljanitor/values.yaml#L113

yonghaoy (approved):

yonghaoy:
Also one random thought just came to my mind. 
Shall we turn on Janitor for Bee's WSM (assume if we are using WSM to create Azure resources)? 
I only see Janitor is on in WSM integration test: https://github.com/DataBiosphere/terra-workspace-manager/blob/6b1af53fd7a0c4a917fef84b2e214374e9cc352a/service/src/test/resources/application-connected-test.yml#L14

src/test/resources/application-integration.yml
rtitle: Yup! I just intended here to update the integration test config (the old MRG was created in 2021).

rtitle:
> Shall we turn on Janitor for Bee's WSM

Janitor is enabled in the [rawls-e2e-azure-tests](https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/workspacemanager/bee/rawls-e2e-azure-tests.yaml#L2) BEE template. That's in fact why I'm making this change -- the Azure integration tests from BEEs are currently leaking a lot of Azure resources. 😅 

Janitor is still disabled in the default `swatomation` BEE template. I think the reason is that people use those BEEs as persistent environments. A flow where Janitor cleanup could be triggered upon BEE deletion would be cool (I feel like I've talked with @jack-r-warren about that in the past..).

jack-r-warren:
Yeah, I'm not really sure how Janitor works so I'm not sure how difficult it would be to include or trigger with BEEs. Sherlock could expose a list of BEE names or namespaces that have been deleted. Not sure how Janitor knows what to target

rtitle:
I think there may need to be Janitor logic changes to support that.

marctalbott (approved):

marctalbott:
The Janitor learns about resources for cleanup from a pub/sub queue that CRL publishes to at creation time. If some process during BEE deletion could publish messages to that queue for all remaining resources in WSM's DB there'd be no Janitor logic changes needed. Don't know how feasible that would be in reality though

yonghaoy:
> The Janitor learns about resources for cleanup from a pub/sub queue that CRL publishes to at creation time. If some process during BEE deletion could publish messages to that queue for all remaining resources in WSM's DB there'd be no Janitor logic changes needed. Don't know how feasible that would be in reality though

If client uses CRL, e.g. WSM, simply add [CleanupConfig](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/crl/CrlService.java#L556) in CrlConfig .
If client does not use CRL, or want to directly publish message to janitor, publish a `ResourceUid` object to Janitor publish queue. ([RBS example](https://github.com/DataBiosphere/terra-resource-buffer/blob/4aa6ba367918811e7c8f163d90130b2701aaa229/src/main/java/bio/terra/buffer/service/cleanup/CleanupScheduler.java#L114C33-L114C33))

yonghaoy:
> > The Janitor learns about resources for cleanup from a pub/sub queue that CRL publishes to at creation time. If some process during BEE deletion could publish messages to that queue for all remaining resources in WSM's DB there'd be no Janitor logic changes needed. Don't know how feasible that would be in reality though
> 
> 1: If client uses CRL, e.g. WSM, simply add [CleanupConfig](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/crl/CrlService.java#L556) in CrlConfig .
> 
> 2: If client does not use CRL, or want to directly publish message to Janitor, publish a `ResourceUid` object to Janitor publish queue. ([RBS example](https://github.com/DataBiosphere/terra-resource-buffer/blob/4aa6ba367918811e7c8f163d90130b2701aaa229/src/main/java/bio/terra/buffer/service/cleanup/CleanupScheduler.java#L114C33-L114C33))

And Janitor even has an endpoint that does the same thing: https://crljanitor.tools.integ.envs.broadinstitute.org/swagger-ui.html#/janitor/createResource
It will create a Resource in Janitor DB then Janitor will schedule to clean it up. (The current AuthZ is not well written and it only takes list of accounts (stored in Vault)).

PR #171, opened 2023-12-08
[PF-2979] Upgrade Janitor to Spring Boot 3

rtitle:
https://broadworkbench.atlassian.net/browse/PF-2979

Used https://github.com/DataBiosphere/terra-resource-buffer/pull/285/ as a model, which is similar to Janitor.

Upgrades Janitor to Spring Boot 3, along with:
* New TCL (including new logback)
* Jakarta client
* OpenTelemetry
* etc

Tested via unit and integration tests. Also ran Janitor locally and manually submitted pubsub messages, and verified resources were deleted.


build.gradle
```diff
@@ -54,63 +63,72 @@ repositories {
...
     constraints {
         implementation group: 'com.google.guava', name: 'guava', version: '31.1-jre' // '-jre' for Java 8 or higher
     }
-    implementation platform('com.google.cloud:libraries-bom:26.0.0') // use common bom
-
-    implementation group: 'com.google.auto.value', name: 'auto-value-annotations', version: '1.9'
-    implementation group: 'com.google.cloud', name: 'google-cloud-billing', version: '2.3.0'
-    implementation group: 'com.google.cloud', name: 'google-cloud-core-http', version: '2.8.0'
-    implementation group: 'com.google.cloud', name: 'google-cloud-pubsub', version: '1.120.0'
-    implementation group: 'com.google.cloud', name: 'google-cloud-storage', version: '2.9.2'
```
rtitle, line 85: Removed some Google dependencies that get included through CRL.

src/main/java/bio/terra/janitor/service/cleanup/MetricsHelper.java
```diff
@@ -2,189 +2,131 @@
...
 import java.time.Duration;
-import java.util.Arrays;
+import org.springframework.stereotype.Component;
 
 /** Helper class for recording metrics associated with cleanup. */
-class MetricsHelper {
-  private MetricsHelper() {}
-
+@Component
+public class MetricsHelper {
```
rtitle, line 15: Refactored `MetricsHelper` to use OpenTelemetry API. Really used RBS as a model.

build.gradle
```diff
@@ -1,3 +1,11 @@
+// Build Script Classpath
+buildscript {
+    ext {
+        springBootVersion = '3.1.2'
```
okotsopoulos, line 4: Might you be able to bring this up further, at least to 3.1.7?  (At that point, Spring Boot has upgraded logback past vulnerable versions.)

build.gradle
```diff
@@ -54,63 +63,72 @@ repositories {
...
-    implementation group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: jerseyVersion
+    implementation group: 'org.glassfish.jersey.inject', name: 'jersey-hk2'
 
     // Kubernetes client
     implementation group: 'io.kubernetes', name: 'client-java', version: '18.0.0'
 
-    // Metrics & Tracing
-    implementation group: 'io.opencensus', name: 'opencensus-exporter-stats-stackdriver', version: '0.31.1'
+    // OpenTelemetry
+    var openTelemetryVersion = '1.31.0'
```
okotsopoulos, line 107:
```suggestion
    var openTelemetryVersion = '1.32.0'
```

(As an aside, I'm wondering whether these imports should be included by default in TCL.)

build.gradle
```diff
@@ -131,27 +149,6 @@ dependencies {
         implementation('org.bouncycastle:bcpkix-jdk15on:1.70')
         implementation('org.bouncycastle:bcprov-jdk15on:1.70')
         spotbugs('org.apache.bcel:bcel:6.6.1')
-        // TODO(PF-2299): Remove these once all dependencies are using >3.0.0. This
```
okotsopoulos, line 134: 🎉

src/main/resources/application.yml
```diff
@@ -56,3 +59,16 @@ spring:
     static-locations: classpath:/api/
   config:
     import: optional:file:config/local-properties.yml
+
+otel:
+  sdk:
+    disabled: false # set to true to disable all open telemetry features
+  traces:
+    sampler:
+      probability: ${SAMPLING_PROBABILITY:0}
```
okotsopoulos, line 68: Linking @dvoet 's correction in the ECM upgrade -- I think sampling config is handled differently.

https://github.com/DataBiosphere/terra-external-credentials-manager/pull/161#discussion_r1426768904

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
```diff
@@ -37,16 +37,14 @@
...
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.result.MockMvcResultHandlers;
 
 @Tag("unit")
 @ActiveProfiles({"test", "unit"})
 @ExtendWith(SpringExtension.class)
-@ContextConfiguration(classes = Main.class)
-@SpringBootTest
+@SpringBootTest(classes = Main.class, properties = "terra.common.prometheus.endpointEnabled=false")
```
okotsopoulos, line 47: Is there a reason that this property override isn't instead put in a test-specific property file?

okotsopoulos (approved):
Looks good to me, but I'm less savvy with the OTEL stuff!

zloery (approved):

build.gradle
rtitle: Aha - yes I will do that. Maybe that means we could also upgrade Spring Boot in TCL and remove the [explicit logback dependency](https://github.com/DataBiosphere/terra-common-lib/pull/127).

build.gradle
rtitle: Hm yeah, it seems TCL does pull them in.. maybe they are not needed here? Will try.

src/main/resources/application.yml
rtitle: Thanks!

src/test/java/bio/terra/janitor/app/controller/JanitorApiControllerTest.java
rtitle: No reason, good idea -- there is an `application-test.yml` file in Janitor.

build.gradle
okotsopoulos: Great idea!

dvoet (approved):
LGTM, the other reviewers are on top of it

build.gradle
rtitle: https://github.com/DataBiosphere/terra-common-lib/pull/130

build.gradle
rtitle: Looks like they do need to be in Janitor because they are marked `implementation` in TCL (so don't get pulled in transitively). I bumped to 1.32.0 though.

build.gradle
rtitle: Bumped Janitor to 3.1.7 (3.2.1 was giving me issues). Will follow up with TCL after.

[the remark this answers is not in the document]
rtitle: Bumped to 3.1.7 (3.2.1 was giving me issues, haven't debugged too much yet).

PR #177, opened 2025-01-10
CORE-250: upgrade Postgres to 16.6

davidangb:
Upgrades unit tests and local development to Postgres 16.6.

The unit tests passed 10 times in a row in GHA for this PR (I re-ran it manually). I do see intermittent unit test failures when running locally - but they're not consistent, and different test cases fail on different runs. I am inclined to write that off as an environmental issue. When running locally, unit tests pass about half the time with only one test causing each failure.

We _could_ go all the way to Postgres 17. However, Postgres 16 is the GCP default (see [here](https://cloud.google.com/sql/docs/postgres/db-versions)) and in my opinion this is not the place to push the envelope, nor would Janitor particularly benefit from the newer version. Postgres 16 does not hit extended support charges until February 1, 2029.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:12.3
+        image: postgres:14.15
         env:
           POSTGRES_PASSWORD: postgres
         options: >-
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:12.3
+        image: postgres:16.6
```
davidangb, line 27: 16.6 is the version supplied by Google for CloudSQL: https://cloud.google.com/sql/docs/postgres/db-versions

kevinpalis (approved):
LGTM!

samanehsan (approved):

PR #190, opened 2025-02-10
CORE-69: update to latest github actions; action cleanup

davidangb:
* Use the latest versions of github actions
* Use `setup-java`'s cache for gradle instead of an explicit call to `actions/cache`
* remove an extraneous/unknown input on `google-github-actions/auth`


dvoet (approved):

kevinpalis (approved):
LGTM

PR #212, opened 2026-05-29
[CTM-511] update netty

calypsomatic:
https://broadworkbench.atlassian.net/browse/CTM-511

Upgrades all io.netty artifacts to 4.1.133.Final to address CVE-2026-42587 using resolutionStrategy.dependencySubstitution rather than resolutionStrategy.force, because azure-core-http-netty:1.15.10 declares netty dependencies with Gradle {strictly} version constraints in its module metadata, which force cannot override in Gradle 8 but substitution can.


aednichols (approved):

davidangb (approved):
kinda scary override, but 👍

